PR #6, opened 2025-10-02
feat(authz): [FC-0099] create LibrariesUserManager view to manage roles for a specific user

dcoa:
### Description 
This PR introduces a second view for managing user roles and permissions. When a team member is selected via the edit button in the table, the user is redirected to this new view to configure their roles and permissions.

>[!warning]
> This PR depends on https://github.com/openedx/frontend-app-admin-console/pull/5 for getting the roles and permissions metadata.

### Changes overview
- Integrate `Link` component in the Layout to manage the `Breadcrumbs` redirections.
- Introduces a reusable `RoleCard` component that displays individual roles in a card layout. Each card includes a collapsible section listing the associated permissions, organized by resource.
    - Permissions are grouped by resource (e.g., Team, Collections).
    - Each permission is displayed as a chip.
    - Chips are enabled or disabled based on the actions allowed by the role.
    - Designed for reuse across views where role-permission mapping needs to be displayed or managed.
- `LibrariesUserManager` view to manage a specific user role assignment and integrate it in the module router.

### Out of scope
- Error handling
- Edit/Delete/Add roles for the current user. 

### Evidence

https://github.com/user-attachments/assets/567e7cd1-a4c5-42dc-8609-098e0235e373

## Testing instructions

1. Create and setup a local environment (Tutor recommended)
2. Run LMS and CMS, then clone the repo in the current branch and run the dev server `npm run dev`
3. Add the URL to the `CORS_WHITELIST` in both LMS/CMS settings.py

```py
# admin console MFE
CORS_ORIGIN_WHITELIST.append("http://apps.local.openedx.io:2025")
LOGIN_REDIRECT_WHITELIST.append("apps.local.openedx.io:2025")
CSRF_TRUSTED_ORIGINS.append("http://apps.local.openedx.io:2025")
```
5. Create a library
6. Navigate to `http://apps.local.openedx.io:2025/admin-console/authz/libraries/:libraryId`
7. Mock the API calls for `/authz/*` queries,  you can use 2 options 
   - Use Chrome override https://developer.chrome.com/docs/devtools/overrides
   - Use an API mock tool such as Postman and Mockoon, and replace the API `baseUrl` with your service and, change `getAuthenticatedHttpClient` to `getHttpClient` to avoid user checks from the backend.

### Additional information
1. https://github.com/orgs/openedx/projects/86/views/1?pane=issue&itemId=128543008&issue=openedx%7Copenedx-authz%7C48

### DIFF
--- a/src/authz-module/components/RoleCard/index.test.tsx
+++ b/src/authz-module/components/RoleCard/index.test.tsx
@@ -1,5 +1,6 @@
-import { fireEvent, screen } from '@testing-library/react';
+import { screen } from '@testing-library/react';
 import { renderWrapper } from '@src/setupTest';
+import userEvent from '@testing-library/user-event';
 import RoleCard from '.';
 
 jest.mock('@openedx/paragon/icons', () => ({
@@ -33,7 +34,8 @@ describe('RoleCard', () => {
     ],
   };
 
-  it('renders all role card sections correctly', () => {
+  it('renders all role card sections correctly', async () => {
+    const user = userEvent.setup();
     renderWrapper(<RoleCard {...defaultProps} />);
 
     // Title
@@ -55,7 +57,7 @@ describe('RoleCard', () => {
     // Collapsible title
     expect(screen.getByText('Permissions')).toBeInTheDocument();
 
-    fireEvent.click(screen.getByText('Permissions'));
+    await user.click(screen.getByText('Permissions'));
 
     // Resource label
     expect(screen.getByText('Library Resource')).toBeInTheDocument();


src/authz-module/components/RoleCard/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+import {
+  Card, Collapsible, Container, Icon, IconButton,
+} from '@openedx/paragon';
+import { Delete, Person } from '@openedx/paragon/icons';
+import PermissionRow from './PermissionsRow';
+import messages from './messages';
+
+interface CardTitleProps {
+  title: string;
+  userCounter?: number | null;
```
arbrandes: I don't see this property being actually passed down, anywhere.  The only instance of RoleCard is in LibrariesUserManager, and it doesn't seem to need `userCounter`.  Will a use for it be added later?

arbrandes (commented):
This looks fine in general.  I had a question and a nit: see below.

More importantly, though, I'm seeing a mismatch between the backend roles/permissions and what the frontend seems to expect.  Notably, the `openedx_authz` backend reports that the library author can `edit_library` (see first screenshot), but the frontend seems to expect `edit_library_content`.  What gives?

Image

vs

Image

src/authz-module/components/RoleCard/index.tsx
dcoa: Yes, this is a reusable component that displays actions when is in the `LibrariesUserManagement` view, and information when is in the `LibrariesTeamManager` view. 

The use of the prop is relevant for a different PR https://github.com/openedx/frontend-app-admin-console/pull/7

src/authz-module/components/RoleCard/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+  objectName?: string | null;
+  description: string;
+  showDelete?: boolean;
+  permissions: any[];
+}
+
+const CardTitle = ({ title, userCounter }: CardTitleProps) => (
+  <div className="d-flex align-items-center">
+    <span className="mr-4 text-primary">{title}</span>
+    {userCounter !== null && userCounter !== undefined && (
```
jacobo-dominguez-wgu: If you just do `userCounter !== null` handles both null and undefined cases due to JavaScripts loose equality comparison, so it makes the code a bit easier to read.

arbrandes:
You need to fix a lint error before we merge.

PR #8, opened 2025-10-02
feat: [FC-0099] add filters and sorting functionality to the team table

bra-i-am:
> [!WARNING]
> This PR depends on https://github.com/openedx/frontend-app-admin-console/pull/5 for getting the roles and permissions metadata.

This pull request introduces filtering and sorting capabilities to the Team Members table. 

### Evidence
[Screencast from 15-10-25 16:32:43.webm](https://github.com/user-attachments/assets/205a7eab-907b-4dc7-84e7-e12c70ad57f0)

### Testing instructions

1. Create and set up a local environment (Tutor recommended)
2. Run LMS and CMS, then clone the repo in the current branch and run the dev server `npm run dev`
3. Add the URL to the CORS_WHITELIST in both LMS/CMS settings.py
``` yml
# admin console MFE
CORS_ORIGIN_WHITELIST.append("http://apps.local.openedx.io:2025")
LOGIN_REDIRECT_WHITELIST.append("apps.local.openedx.io:2025")
CSRF_TRUSTED_ORIGINS.append("http://apps.local.openedx.io:2025")
```
4. Go to the Studio and create a library

6. Install the plugin https://github.com/openedx/openedx-authz by running `tutor dev exec lms pip install git+https://github.com/openedx/openedx-authz`
7. Run the migrations `tutor dev exec lms python manage.py lms migrate openedx_authz`
8. Load the default policies by running `tutor dev exec lms python manage.py lms load_policies`
9. Go to http://local.openedx.io:8000/api-docs/#/authz/authz_v1_roles_users_update and add roles to some users, following the structure
```
{
  "role": "library_author", // Create another with "library_admin" to be able to filter
  "scope": "lib:OpenedX:testlib", // Id of the previously created library
  "users": [
    "admin"
    // more usernames or emails if wanted
  ]
}
```
10. Navigate to http://apps.local.openedx.io:2025/admin-console/authz/libraries/< your_library_id > and modify the filters; data should be properly filtered and sorted

### Additional Information
- https://github.com/openedx/openedx-authz/issues/50

### DIFF
--- a/src/authz-module/libraries-manager/components/TeamTable/components/SortDropdown.tsx
+++ b/src/authz-module/libraries-manager/components/TeamTable/components/SortDropdown.tsx
@@ -35,8 +35,7 @@ const SortDropdown: FC = () => {
   const SORT_LABELS: Record<string, string> = useMemo(() => ({
     'name-a-z': intl.formatMessage({ id: 'authz.libraries.team.table.sort.name-a-z', defaultMessage: 'Name A-Z' }),
     'name-z-a': intl.formatMessage({ id: 'authz.libraries.team.table.sort.name-z-a', defaultMessage: 'Name Z-A' }),
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }), []);
+  }), [intl]);
 
   const currentSort = useMemo(() => {
     if (!state?.sortBy?.length) { return undefined; }
@@ -63,8 +62,7 @@ const SortDropdown: FC = () => {
       ...option,
       label: SORT_LABELS[key],
     })),
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-    [],
+    [SORT_LABELS],
   );
 
   const currentSortLabel = sortOrder ? SORT_LABELS[sortOrder] : 'Sort';
--- a/src/authz-module/libraries-manager/components/TeamTable/index.tsx
+++ b/src/authz-module/libraries-manager/components/TeamTable/index.tsx
@@ -1,4 +1,4 @@
-import { useMemo } from 'react';
+import { useEffect, useMemo } from 'react';
 import { useNavigate } from 'react-router-dom';
 import debounce from 'lodash.debounce';
 import { useIntl } from '@edx/frontend-platform/i18n';
@@ -80,6 +80,10 @@ const TeamTable = () => {
     [roles],
   );
 
+  const fetchData = useMemo(() => debounce(handleTableFetch, 500), [handleTableFetch]);
+
+  useEffect(() => () => fetchData.cancel(), [fetchData]);
+
   return (
     <DataTable
       isFilterable
@@ -90,7 +94,7 @@ const TeamTable = () => {
       manualSortBy
       defaultColumnValues={{ Filter: TextFilter }}
       numBreakoutFilters={3}
-      fetchData={debounce(handleTableFetch, 1000)}
+      fetchData={fetchData}
       data={rows}
       itemCount={teamMembers?.count || 0}
       pageCount={pageCount}


jacobo-dominguez-wgu (commented):
I tested the filters by username/email, role sorting, and alphabetical sorting — everything seems to be working well!

That said, I noticed a small opportunity to improve the user experience. There’s a slight delay before the skeleton appears when changing the alphabetical sorting, filtering by role or username/email, and when clicking the “Clear filters” button. Occasionally, the skeleton doesn’t appear at all. I’ve attached a short video showing what I observed:

https://github.com/user-attachments/assets/8773280d-cb53-41c5-824a-c9233ccda8b1

Also, I have a quick question about a potential edge case:
if the admin wants to find and remove all the collaborator roles from an specific email domain so he set the filtering for that,  after filters are applied he clicks on edit button for the first row and then removes the team member role, then returns to the table and the filters seem to reset. Is that the expected behavior, or should the filtering persist after returning?

src/authz-module/data/api.ts
```diff
@@ -24,9 +33,25 @@ export interface AssignTeamMembersRoleRequest {
   scope: string;
 }
 
-export const getTeamMembers = async (object: string): Promise<TeamMember[]> => {
-  const { data } = await getAuthenticatedHttpClient().get(getApiUrl(`/api/authz/v1/roles/users/?scope=${object}`));
-  return camelCaseObject(data.results);
+// TODO: replece api path once is created
```
brian-smith-tcril: Could you elaborate on this TODO? What will the API path be replaced with?

src/authz-module/data/hooks.ts
```diff
@@ -20,17 +21,19 @@ const authzQueryKeys = {
...
  * ```
  */
-export const useTeamMembers = (object: string) => useQuery<TeamMember[], Error>({
-  queryKey: authzQueryKeys.teamMembers(object),
-  queryFn: () => getTeamMembers(object),
+export const useTeamMembers = (scope: string, querySettings: QuerySettings) => useQuery<GetTeamMembersResponse, Error>({
+  queryKey: authzQueryKeys.teamMembers(scope, querySettings),
+  queryFn: () => getTeamMembers(scope, querySettings),
   staleTime: 1000 * 60 * 30, // refetch after 30 minutes
+  refetchOnWindowFocus: false,
```
brian-smith-tcril: Just curious as to why this is being added. It doesn't seem like something that would be a sort/filter/search specific change.

src/authz-module/libraries-manager/components/TeamTable/components/SortDropdown.tsx
```diff
@@ -0,0 +1,96 @@
...
+  id: string;
+  desc: boolean;
+  label: string;
+}
+
+interface SortByOptions {
+  [key: string]: Omit<SortOption, 'label'>;
+}
+
+const SORT_BY_OPTIONS: SortByOptions = {
```
brian-smith-tcril: Not blocking, but it'd be great to update this component to take in the `SortByOptions` and `SORT_LABELS` as props so it can be reused.

src/authz-module/libraries-manager/components/TeamTable/index.tsx
```diff
@@ -1,12 +1,19 @@
+import { useMemo } from 'react';
 import { useNavigate } from 'react-router-dom';
+import debounce from 'lodash.debounce';
```
brian-smith-tcril: I'm curious about the need to debounce this here. I'd think `react-query` would have built-in ways to handle this nicely.

I guess my main concern here is that debouncing is addressing a symptom instead of figuring out the root of the issue.

`frontend-app-authoring` has a similar `DataTable` backed by `react-query` and isn't using a debounce, so it might be worth looking into how it's being handled there?

https://github.com/openedx/frontend-app-authoring/blob/82a3c2815bc776a3aa23dd82bbd8848441e12123/src/taxonomy/tag-list/TagListTable.jsx#L71-L77

https://github.com/openedx/frontend-app-authoring/blob/82a3c2815bc776a3aa23dd82bbd8848441e12123/src/taxonomy/data/apiHooks.ts#L179-L191

brian-smith-tcril (commented):
Overall this looks great!

The components being added here look like they'd likely be useful to have in more places than just this MFE. It'd be awesome to see what it'd take to get them added to Paragon.

I left a few comments with questions.

Thank you so much for this PR!

src/authz-module/libraries-manager/components/TeamTable/index.tsx
dcoa: The reasoning behind the `debounce`  is leverage better performance for the search input mainly, and reduce the number of API calls with the user typing. As I can see the example in authoring is only managing pagination and that is why is not needed there.

Currently, there is not a built-in solution in `react-query` as you can see here https://github.com/TanStack/query/discussions/8423 (there are working on a second library to support it but is still in alpha release)

brian-smith-tcril: From https://github.com/TanStack/query/discussions/8423#discussioncomment-11517924

> 4. Debouncing itself is likely a workaround
> 
> For example, for auto-complete, we likely just don’t want to send events until the user has “stopped” typing. Defining an arbitrary time for what “stopped” means seems ... arbitrary? In my example, it will also delay the fetch by 500ms after the user has stopped.
> 
> Since React 18, React offers the built-in `useDeferredValue` hook for low-prio updates that can be interrupted by high-prio updates (e.g. user input). With the user-land solution, you can do:
> 
> ```tsx
> const deferredSearchTerm = React.useDeferredValue(searchTerm)
> 
> const { data } = useQuery({
>  queryKey: ['search', deferredSearchTerm]
>  queryFn: fetchSearchResults,
> });
> ```
> 
> and kind of get the best of both worlds with a react built-in feature.

I'm fine with keeping debounce for this PR, but it'd be good to make a follow-up issue to see if we can replace it with [`useDeferredValue`](https://react.dev/reference/react/useDeferredValue).

brian-smith-tcril (approved):
LGTM!

I'm still curious about https://github.com/openedx/frontend-app-admin-console/pull/8#discussion_r2452781418, but I don't feel like it's *hurting* anything (I'd really only be opposed to having `refetchOnWindowFocus` set to `false` for a page where we expect content to be updated rather frequently.

PR #10, opened 2025-10-13
fix(authz): [FC-0099] update roles label in `TeamTable`

dcoa:
# Fix roles label in team table

This Pr update the chip component in the Team Table to display the name of the roles based on the metadata defined here https://github.com/openedx/frontend-app-admin-console/pull/5

**Before**
image

**After**

image


dcoa:
@arbrandes @jacobo-dominguez-wgu this is a small fix

brian-smith-tcril (approved):
LGTM!

Tested locally and the role names are appearing nicely!

PR #11, opened 2025-10-13
feat(authz): [FC-0099] create an custom error boundery

dcoa:
# Create a custom error boundary for libraries team management workflow

This PR creates a custom error boundary that maps errors `401/403/404/5xx`

## Changes overview

- Add a env variable to performs redirections to authoring MFE libraries
- Create a reload button for 5xx error that will retry the query
- Create a custom error to display error access denied when the user do not have permissions over a library (the validation permission endpoint does not performs that)

## Evidence

image

image

### DIFF
--- a/src/authz-module/libraries-manager/ErrorPage/index.test.tsx
+++ b/src/authz-module/libraries-manager/ErrorPage/index.test.tsx
@@ -1,6 +1,5 @@
-import {
-  screen, fireEvent, waitFor,
-} from '@testing-library/react';
+import { screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
 import { ErrorBoundary } from 'react-error-boundary';
 import { renderWrapper } from '@src/setupTest';
 import LibrariesErrorFallback from './index';
@@ -67,9 +66,8 @@ describe('LibrariesErrorFallback', () => {
         <ThrowError error={error} />
       </ErrorBoundary>,
     );
-    const reloadBtn = screen.getByText(/Reload Page/i);
-    fireEvent.click(reloadBtn);
-    // If your ErrorPage uses error.refetch, this will be called
-    await waitFor(() => expect(refetch).toHaveBeenCalled());
+    const user = userEvent.setup();
+    await user.click(screen.getByText(/Reload Page/i));
+    expect(refetch).toHaveBeenCalled();
   });
 });


src/authz-module/libraries-manager/ErrorPage/index.tsx
```diff
@@ -0,0 +1,95 @@
...
+    title, description, statusCode, showBackButton, showReloadButton,
+  } = getErrorConfig({ errorMessage, errorStatus });
+
+  const handleReload = () => {
+    if (reloading) { return; }
+    setReloading(true);
+    resetErrorBoundary();
+  };
+  return (
+    <Container className="d-flex flex-column align-items-center justify-content-center min-vh-100 bg-light-200" data-testid="error-page">
```
arbrandes:
```suggestion
    <Container className="d-flex flex-column align-items-center justify-content-center min-vh-100 bg-light-200">
```

Do you foresee a need to use `getByTestId`?  If not, I'd rather we only add it if necessary ([for the usual reasons](https://testing-library.com/docs/queries/about/#priority)).

arbrandes (changes requested):
Looks good, though I did have one nit.

src/authz-module/libraries-manager/ErrorPage/index.tsx
```diff
@@ -0,0 +1,95 @@
...
+  }
+  return ({
+    title: messages['error.page.title.generic'],
+    description: messages['error.page.message.generic'],
+    showBackButton: true,
+    showReloadButton: true,
+  });
+};
+
+const librariesUrl = () => `${getConfig().AUTHORING_BASE_URL}/authoring/libraries`;
```
jacobo-dominguez-wgu: I am not getting why this one is a function and not just a constant.
```suggestion
const LIBRARIES_URL = `${getConfig().AUTHORING_BASE_URL}/authoring/libraries`;
```
And maybe it can be moved to the constants file.

src/authz-module/libraries-manager/ErrorPage/index.tsx
```diff
@@ -0,0 +1,95 @@
...
+  const [reloading, setReloading] = useState(false);
+
+  const errorStatus: number = error?.customAttributes?.httpErrorStatus;
+  const errorMessage: string = error?.message;
+  const {
+    title, description, statusCode, showBackButton, showReloadButton,
+  } = getErrorConfig({ errorMessage, errorStatus });
+
+  const handleReload = () => {
+    if (reloading) { return; }
```
jacobo-dominguez-wgu: This line maybe not needed since the button is disabled when reloading is true.

jacobo-dominguez-wgu (commented):
I have not tested on my local yet, but overall it looks good. I have added a couple of comments with some suggestions let me know your thoughts. Nice work!

dcoa:
@jacobo-dominguez-wgu @arbrandes  I addressed your comments, and I just change the env variable to `COURSE_AUTHORING_MICROFRONTEND_URL` to avoid duplicate values in tutor-mfe for example.

dcoa:
@jacobo-dominguez-wgu we have a sandbox now https://openedx.atlassian.net/wiki/spaces/OEPM/pages/5279023110/AuthZ+Stage+Information

arbrandes (approved):

PR #24, opened 2025-10-23
fix: [FC-0099] modify app name in the html file

dcoa:
This PR modifies the name of the application from template to admin console.


jacobo-dominguez-wgu (approved):
Tested on my local and title looks fine!
image

arbrandes (approved):

Issue #29, opened 2025-10-27
Bug: Breadcrumb “Manage Access” does not navigate, reloads same screen


gviedma-aulasneo:
**Current behavior**
Clicking the “Manage Access” breadcrumb redirects to the same screen the user is already on, regardless of the section.

**Expected behavior**
Clicking “Manage Access” should take the user to the library default view:
admin-console/authz/libraries/$currentLibrary

**Steps to reproduce**
1) Log in with a user who has Studio access and the “manage team members” permission over the library.  
2) Open a team member’s user detail view, note that only third party users are eligible.  
3) Click the “Manage Access” breadcrumb.

dcoa:
@jacobo-dominguez-wgu could you help me with a backport of this for the ulmo branch? 

**Note:** from the current list of issues listed in the repo, this is the only one that need to be backported.

PR #32, opened 2025-11-04
fixing base breadcrumb item with no link set

jacobo-dominguez-wgu:
## Description
Adding url link to “Manage Access” breadcrumb item and adding unit test to support this modification.
It closes https://github.com/openedx/frontend-app-admin-console/issues/29

### Demo video
https://github.com/user-attachments/assets/5168099d-0b68-48c1-b0e0-8a9d511bb8ad

### How to test it
- `npm install && npm run dev`
- Go to `http://apps.local.openedx.io:2025/admin-console/authz/libraries/<YOUR_LIB_ID>`
- Click on the `Edit` link on one of the users in the team members table or go to `http://apps.local.openedx.io:2025/admin-console/authz/libraries/<YOUR_LIB_ID>/<username>`
- Click on the `Manage Access` item on the breadcrumb.
- It must take you back to the `http://apps.local.openedx.io:2025/admin-console/authz/libraries/<YOUR_LIB_ID>` page

### DIFF
--- a/src/authz-module/index.scss
+++ b/src/authz-module/index.scss
@@ -3,11 +3,6 @@
 .authz-libraries {
   --height-action-divider: 30px;
 
-  .pgn__breadcrumb li:first-child a {
-    color: var(--pgn-color-breadcrumb-active);
-    text-decoration: none;
-  }
-
   hr {
     border-top: var(--pgn-size-border-width) solid var(--pgn-color-border);
     width: 100%;
--- a/src/authz-module/libraries-manager/LibrariesTeamManager.test.tsx
+++ b/src/authz-module/libraries-manager/LibrariesTeamManager.test.tsx
@@ -191,4 +191,12 @@ describe('LibrariesTeamManager', () => {
     expect(readPublicToggle).toBeInTheDocument();
     expect(readPublicToggle).toBeDisabled();
   });
+
+  it('renders correct navigation link label and URL on breadcrumb', () => {
+    renderTeamManager();
+    const navLink = screen.getByRole('link', { name: 'Manage Access' });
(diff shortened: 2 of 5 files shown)


dcoa:
> Hi @dcoa, I noticed there are some [custom styles](https://github.com/openedx/frontend-app-admin-console/blob/master/src/authz-module/index.scss#L6) overriding the defaults for the first item on the breadcrumb, I did not find the figma design for the breadcrumb. What's the reason for this style override?

As you can see in the original design, the "Manage Access" is always displayed as a static (no-actionable item).

image

image

The intention overriding the styles was simulate that static behavior (because there was no page to redirect)

Now, I feel that if a link needs to be added then the default styles should be displayed to keep the look and feel of the website.

jacobo-dominguez-wgu:
> > Hi @dcoa, I noticed there are some [custom styles](https://github.com/openedx/frontend-app-admin-console/blob/master/src/authz-module/index.scss#L6) overriding the defaults for the first item on the breadcrumb, I did not find the figma design for the breadcrumb. What's the reason for this style override?
> 
> As you can see in the original design, the "Manage Access" is always displayed as a static (no-actionable item).
> 
> image image
> The intention overriding the styles was simulate that static behavior (because there was no page to redirect)
> 
> Now, I feel that if a link needs to be added then the default styles should be displayed to keep the look and feel of the website.

Sure now that it has a url it make sense to look like a normal link , I am removing those custom styles so the breadcrumb link can have the defaults from paragon, thanks.

dcoa (approved):
LGTM! Thank you!!

@brian-smith-tcril @arbrandes could you please have a look on this small fix.

Issue #14, opened 2025-10-21
Better manage in the feedback for user input in add team member modal


dcoa:
### Context: 

Related link https://github.com/openedx/frontend-app-admin-console/pull/3#issuecomment-3426769405

For the multiple user input, the input need a better way to display the error users when the save action is trigger and the `textarea` modified.

Currently the modal validates the input as follows:

- Add a user or multiple emails or usernames comma separated
- If the user exist and the role assignment was successful then close the modal.
- If there is a multi user assignment and was partially completed, the users who pass are removed from the input and the error uses are displayed and marked in red.

During the development the input was manage to validate if the user modify the input and do not introduce the previous error return the color to black, but if the some of the error users are in the `textarea` this keeps the red color.

As is displayed here https://github.com/openedx/frontend-app-admin-console/pull/3#issuecomment-3427110869

The idea here is think in a better way to display the error uses and their validation if possible.

MaferMazu:
I put the label bug because the behavior @BryanttV mentioned causes the modal not to close as it should.

cc @dcoa

MaferMazu:
@jacobo-dominguez-wgu my general thought is it sounds good, only make sure we are following the [accessibility best practices](https://docs.openedx.org/en/latest/developers/concepts/accessibility.html).

PR #42, opened 2025-11-20
fix: changing how the feedback message is displayed for adding team members

jacobo-dominguez-wgu:
## Description
This PR is intended to improve the feedback process and have a better feedback system for user's input in the add team member modal.
Closes https://github.com/openedx/frontend-app-admin-console/issues/14.

### **What it changes:**

- **Improved Toast Feedback System**
**Before**: Single generic toast message for all scenarios
**After**: Separate toast messages for each status type:
✅ Successfully added users
❌ User already has the assigned role
❌ User not found
❌ Generic errors
This provides clearer, more actionable feedback to users.

- **Enhanced Message Clarity**
**Added user identification**: Toast messages now include the specific userId that triggered each status
**Example**: `The user already has the role (jon@mail.com)`
**Benefit**: Users can easily identify which specific user caused each message when multiple toasts are displayed

- **Visual Error Indicators**
**Extended red text functionality**: Previously only "user not found" errors showed red text in the input textarea
**Now applies to all error types**: Any user causing an error will appear in red in the textarea
**Improved UX**: Users can quickly identify problematic entries and fix them

- **Better Error Handling Flow**
**Modal behavior**: The "Add Team Member" modal no longer auto-closes when errors occur
**Why**: Allows users to immediately correct typos or invalid entries without reopening the modal
**Result**: Smoother error recovery workflow

- **Dynamic Toast Duration** 
**Adaptive timing**: Toast display duration now scales with the number of messages
**Formula**: 5 seconds × number of toast messages displayed
**Rationale**: Gives users adequate time to read multiple concurrent messages (up to 4 different toast types)

### Demo video

https://github.com/user-attachments/assets/3bf09d9e-b713-4adb-8188-7cca3a53d205

### How to test it
- `npm install && npm run dev`
- Go to http://apps.local.openedx.io:2025/admin-console/authz/libraries/<YOUR_LIB_ID>
- Click on the add team member button to continue with the following scenarios:

We can test all the error cases described [here](https://github.com/openedx/frontend-app-admin-console/issues/14#issuecomment-3554932899), but also the next case scenarios covers most of the error and success cases:
#### Test scenario 1 (success)
Add a role to an existing user which role is not already set.
Expected: It should show the success toast message for 5 seconds and the modal automatically closes.

#### Test scenario 2 (already set error)
Add a role to an existing user which role is already set.
Expected: It should show the already set toast message for 5 seconds ~~and the modal does not close~~.

#### Test scenario 3 (two usernames, one successful and one error)
Type the username of an existing user and one nonexisting and asign a role not already set.
Expected: It should show the success toast message and the non existing user message for 10 seconds. The modal does not closes and the username of the nonexisting and input text area should be in error state.

#### Test scenario 4 (three usernames, one successful and two errors)
Type the username of an existing user, one nonexisting user and one existing user with a role asigned (user3), asign the role of the user 3.
Expected: It should show the success toast message, the non existing user message and the already set role error message for 15 seconds. The modal does not closes and the usernames of the nonexisting user and the already set user should be in error state.

### DIFF
--- a/src/authz-module/libraries-manager/components/AddNewTeamMemberModal/messages.ts
+++ b/src/authz-module/libraries-manager/components/AddNewTeamMemberModal/messages.ts
@@ -58,7 +58,7 @@ const messages = defineMessages({
   },
   'libraries.authz.manage.add.member.failure.not.found': {
     id: 'libraries.authz.manage.add.member.failure.not.found',
-    defaultMessage: '<Bold>We couldn\'t find a user for {count, plural, one {# email address or username} other {# email addresses or usernames}} ({userIds}).</Bold><Br></Br> Please check the values and try again, or invite them to join your organization first.',
+    defaultMessage: '<Bold>We couldn\'t find a user for {count, plural, one {# email address or username} other {# email addresses or usernames}}.</Bold><Br></Br> Please check the values ({userIds}) and try again, or invite them to join your organization first.',
     description: 'Error message in case of user not found when adding new team members',
   },
   'libraries.authz.manage.add.member.failure.generic': {
--- a/src/authz-module/libraries-manager/components/AddNewTeamMemberModal/AddNewTeamMemberTrigger.test.tsx
+++ b/src/authz-module/libraries-manager/components/AddNewTeamMemberModal/AddNewTeamMemberTrigger.test.tsx
@@ -272,8 +272,8 @@ describe('AddNewTeamMemberTrigger', () => {
     });
 
     await waitFor(() => {
-      expect(screen.getByText(/We couldn't find a user for 1 email address or username \(unknown@example.com\)/)).toBeInTheDocument();
-      expect(screen.getByText(/The user already has the role \(already@example.com\)/)).toBeInTheDocument();
+      expect(screen.getByText(/We couldn't find a user for 1 email address or username/)).toBeInTheDocument();
+      expect(screen.getByText(/The user already has the role/)).toBeInTheDocument();
     });
 
     expect(screen.getByRole('dialog', { name: 'Add New Team Member' })).toBeInTheDocument();


jacobo-dominguez-wgu:
Backport PR: https://github.com/openedx/frontend-app-admin-console/pull/44

MaferMazu (commented):
Thanks for these improvements @jacobo-dominguez-wgu.

I just have a comment:
Regarding the USER_ALREADY_HAS_ROLE, although the user is not added to the team, the user's experience is that they achieve the goal (the users are in the group). So I think we should consider the case of already having a role as a success, like an informative success.
That implies we shouldn't keep those user/email in red, and if I only have users with the role, we should close the modal. (this is related to the first case in this comment https://github.com/openedx/frontend-app-admin-console/issues/14#issuecomment-3548963215)

What do you think, @jacobo-dominguez-wgu?

The rest looks good to me. I tested, and it works as expected! ✨

jacobo-dominguez-wgu:
> Thanks for these improvements @jacobo-dominguez-wgu.
> 
> I just have a comment: Regarding the USER_ALREADY_HAS_ROLE, although the user is not added to the team, the user's experience is that they achieve the goal (the users are in the group). So I think we should consider the case of already having a role as a success, like an informative success. That implies we shouldn't keep those user/email in red, and if I only have users with the role, we should close the modal. (this is related to the first case in this comment [#14 (comment)](https://github.com/openedx/frontend-app-admin-console/issues/14#issuecomment-3548963215))
> 
> What do you think, @jacobo-dominguez-wgu?
> 
> The rest looks good to me. I tested, and it works as expected! ✨

Hi @MaferMazu I have updated the code to consider the USER_ALREADY_HAS_ROLE cases as a successful message.
This includes, the userIds will not appear in read and will be removed from the textarea (in cases where there are other userIds with errors) and will automatically closes the modal (in cases where there are other no errors), always displaying a toast indicating the already existing role. Can you take a second look to review it?

src/authz-module/libraries-manager/components/AddNewTeamMemberModal/messages.ts
```diff
@@ -53,22 +53,22 @@ const messages = defineMessages({
...
+    defaultMessage: '{count, plural, one {# team member added successfully} other {# team members added successfully}} ({userIds})',
     description: 'Success message when adding new team members',
   },
-  'libraries.authz.manage.add.member.failure': {
-    id: 'libraries.authz.manage.add.member.failure',
-    defaultMessage: '<Bold>We couldn\'t find a user for {count, plural, one {# email address or username.} other {# email addresses or usernames.}}</Bold><Br></Br> Please check the values and try again, or invite them to join your organization first.',
-    description: 'Error message when adding new team members',
+  'libraries.authz.manage.add.member.failure.not.found': {
+    id: 'libraries.authz.manage.add.member.failure.not.found',
+    defaultMessage: '<Bold>We couldn\'t find a user for {count, plural, one {# email address or username} other {# email addresses or usernames}} ({userIds}).</Bold><Br></Br> Please check the values and try again, or invite them to join your organization first.',
```
dcoa: This is a comment not blocking, but I think the user identifier should be part of the body of the notification, like: 

```suggestion
    defaultMessage: '<Bold>We couldn\'t find a user for {count, plural, one {# email address or username} other {# email addresses or usernames}}.</Bold><Br></Br> Please check the values ({userIds}) and try again, or invite them to join your organization first.',
```

dcoa (approved):
I haven't tested in my local but the code seems good to me, thank you!!

MaferMazu:
@arbrandes @brian-smith-tcril, could you help us with a review/merge powers with this PR?

PR #109, opened 2026-03-25
feat(authz): add step 1 for the role assignment wizard

### LINKED ISSUE #91: Task - RBAC AuthZ - US: M2.8 Assign role wizard - Step 1 usernames and role
## Description
Create the step 1 for the assign role wizard which includes the inputs for usernames / emails and the role selection.

Image

### General
- The "Assign Role" CTA is always visible in the console, both in the Team Members tab and in the Roles and Permissions tab.
- When accessing the wizard from the user audit view, the user input in Step 1 is pre-populated with that user.
- The wizard has a breadcrumb that exits the flow and returns to the previous view at any point.
- A Cancel button is available at the bottom of both steps. Clicking it returns to the previous view.

### Step 1 — Who and Role
- Step 1 is titled "Who and Role."
- There is a text input to add one or more usernames or emails, separated by commas.
- Users must have an existing account. If any user in the batch does not exist, the input shows an error in red for the invalid entries and the flow is blocked until all entries are valid.
- Below the input there is a role selector using radio buttons. Only one role can be selected at a time. Selecting a new role deselects the previous one.
- Roles are grouped by courses and libraries. A user only sees the groups they have permissions to assign. If they have no library scopes, they do not see libraries. If they have no course scopes, they do not see courses.
- Course Editor and Course Auditor are grayed out and not selectable.
- There is a link to documentation for roles not available in the list.
- Clicking Next validates the users. If all users exist, the wizard moves to Step 2. If any do not exist, the flow is blocked.

## Design
https://www.figma.com/design/onU2END2OXaF7RRLWEHsZI/AuthZ---v2?node-id=7639-19185&t=FLJ7Zz9Qhu9MtJVZ-4

bra-i-am:
> [!CAUTION]
> This PR requires the validation endpoint introduced in the PR https://github.com/openedx/openedx-authz/pull/245

This PR introduces Step 1 of the Role Assignment Wizard, replacing the previous modal-based approach for adding team members.

### What's included:

- A new `/assign-role` route with a two-step `Stepper` component (AssignRoleWizardPage + AssignRoleWizard)
- Step 1 — Who and Role: users can enter one or more usernames/emails (comma-separated) and select a role. Invalid users are highlighted inline in red via `HighlightedUsersInput`, a custom textarea with a synchronized overlay layer
- User validation via a new `POST /api/authz/v1/users/validate` endpoint before advancing to Step 2
- Role options are filtered based on the current user's actual permissions (MANAGE_LIBRARY_TEAM / MANAGE_COURSE_TEAM), with course roles visible but disabled pending a future update
- Step 2 — Where it applies is scaffolded but not yet implemented; it will define the application scope in a follow-up PR
- Constants for library and course roles/permissions are centralized in `authz-module/constants.ts` and re-exported where needed
- The "Assign Role" button in `LibrariesTeamManager` now navigates to the wizard instead of opening a modal
Not in scope for this PR: Step 2 implementation, course role assignment.

### Additional Information
Resolves https://github.com/openedx/frontend-app-admin-console/issues/91

### Screenshots
image

##### 1. Open the wizard from the home page

1. Navigate to `/authz`.
2. Click **Assign Role** in the top-right header.
3. Confirm you land on `/authz/assign-role` and the wizard Step 1 is shown.

##### 2. Open the wizard pre-filled from the user audit view

1. Navigate to a library team page (`/authz/libraries/<lib-id>`).
2. Click **Assign Role** in the header → confirm the wizard opens at `/authz/assign-role` with the users input **empty** and `from=` pointing back to the team page.
3. Go back, click on any team member `edit` action to open their audit page (`/authz/libraries/<lib-id>/<username>`).
4. Click **Assign Role** in the header → confirm the wizard opens with that team member's username already filled in the users input.

##### 3. Role selection

1. Confirm roles are grouped under **Courses** and **Libraries**.
2. Select a library role, then click a course role — confirm only one radio is checked at a time.
3. Confirm **Course Editor** and **Course Auditor** are visible but disabled (greyed out with a tooltip).

##### 4. Users input — invalid user

1. Type a username that does not exist, select any role, click **Next**.
2. Confirm the invalid username is highlighted in red inside the input.
3. Confirm the message "X username(s) not associated with an account" appears below.
4. Edit the input → confirm the red highlight and error message both disappear immediately.

##### 5. Advance to Step 2

1. Type a valid username, select a role, and click **Next**.
2. Confirm the wizard advances to Step 2.
3. Click **Back** → confirm you return to Step 1 with the username and role still selected.

##### 6. Cancel and breadcrumb return to the originating view

1. Open the wizard from `/authz`, click **Cancel** → confirm you land on `/authz`.
2. Open the wizard from a user audit page, click **Cancel** → confirm you return to that user audit page, not `/authz`.
3. Repeat step 2 but click the breadcrumb link instead of Cancel.

### AssignRoleWizard — Step 1 Test Coverage

| Status | Test |
|---|---|
| ✅ | `Cancel returns to the previous view` |
| ✅ | `opens with the user pre-populated when provided via initialUsers` |
| ❌ | `shows only course roles when user only has manage_course_team permission` |
| ❌ | `shows only library roles when user only has manage_library_team permission` |
| ✅ | `shows both course and library roles when user has both permissions` |
| ✅ | `selecting a different role replaces the previous selection` |
| ✅ | `blocks progression and highlights the unknown user` |
| ✅ | `advances to Step 2 when all users are valid` |
| ✅ | `shows a network error and blocks progression when the validation call fails` |
| ✅ | `clears the invalid user highlights when the input is edited` |


openedx-webhooks:
Thanks for the pull request, @bra-i-am!


</details>

- Dependencies
- Blockers
- Timeline information
- Partner information
- Supporting documentation
</details>





</details><details>


---

</details>

- The need for product review

</details>

arbrandes:
What is the PR (or PRs) that introduces the required endpoints?  Has it already merged?

bra-i-am:
@arbrandes, sorry for not including it in the cover letter earlier. The PR is https://github.com/openedx/openedx-authz/pull/245 and has already been merged.

dcoa:
As a minor comment in the UI can we keep the space between the text area and the feedback?

### The design

image


### Currently 

image


Also the background for the form section in the current implementation is white but the design presented in the issue is light -  Nothing blocking but I mention it just in case.

src/authz-module/wizard/messages.ts
```diff
@@ -0,0 +1,139 @@
...
+    description: 'Placeholder text for the users textarea',
+  },
+  'wizard.step1.users.hint': {
+    id: 'wizard.step1.users.hint',
+    defaultMessage: 'The user must already have an account.',
+    description: 'Hint text below the users textarea',
+  },
+  'wizard.step1.users.invalid': {
+    id: 'wizard.step1.users.invalid',
+    defaultMessage: '{count, plural, one {This email is not associated with an account in this platform.} other {Some of these emails or usernames are not associated with accounts in this platform.}}',
```
dcoa:
```suggestion
    defaultMessage: '{count, plural, one {This email/username is not associated with an account in this platform.} other {Some of these emails or usernames are not associated with accounts in this platform.}}',
```

src/authz-module/data/api.test.ts
```diff
@@ -0,0 +1,173 @@
+import { getAuthenticatedHttpClient } from '@edx/frontend-platform/auth';
```
dcoa: I'm not against introducing this test but I don't think it will introduce much value either. 

I explain better, the functions usually call `getAuthenticatedHttpClient` and return a camel cased value, the logic is simple.
In addition, the hooks test follows the same pattern as here, mocking  `getAuthenticatedHttpClient`, with the idea to cover the `api` during the test coverage. because of that, I prefer modify them if we find improvements than creating a solo api test suite.

src/authz-module/wizard/DefineApplicationScopeStep.tsx
```diff
@@ -0,0 +1,17 @@
...
+  selectedScopes: Set<string>;
+  onScopeToggle: (scopeId: string) => void;
+}
+
+const DefineApplicationScopeStep = (
+  { selectedRole, selectedScopes, onScopeToggle }: DefineApplicationScopeStepProps,
+) => {
+  // Placeholder for Step 2 - "Define Application Scope"
+  // This will be implemented in future iterations
+  console.log(selectedRole, selectedScopes, onScopeToggle);
```
dcoa:
```suggestion

```

src/authz-module/wizard/HighlightedUsersInput.tsx
```diff
@@ -0,0 +1,120 @@
...
+      const key = offset;
+      offset += part.length;
+      if (part === ',') { return <span key={`comma-${key}`}>,</span>; }
+      const trimmed = part.trim();
+      const isInvalid = trimmed.length > 0 && invalidSet.has(trimmed);
+      return (
+        <span
+          key={`part-${key}`}
+          data-invalid={isInvalid || undefined}
+          style={{
```
dcoa:
```suggestion
          className={ isInvalid ? 'text-danger' : ''}
```
Please avoid the use of inline styles, you can use classes instead.

src/authz-module/wizard/AssignRoleWizard.test.tsx
```diff
@@ -0,0 +1,196 @@
+import React from 'react';
+import { screen, waitFor } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import { renderWrapper } from '@src/setupTest';
+import { useValidateUsers } from '../data/hooks';
+import { useValidateUserPermissions } from '../../data/hooks';
+import AssignRoleWizard from './AssignRoleWizard';
+
+// Mock Paragon Stepper — exposes activeKey as data-active-step for step navigation assertions
+jest.mock('@openedx/paragon', () => {
```
dcoa: Please remove all the mocked components (avoid this pattern).

src/authz-module/wizard/AssignRoleWizard.test.tsx
```diff
@@ -0,0 +1,196 @@
...
+    </button>
+  ),
+}));
+
+jest.mock('../data/hooks', () => ({
+  useValidateUsers: jest.fn(),
+}));
+
+jest.mock('../../data/hooks', () => ({
+  useValidateUserPermissions: jest.fn(),
```
dcoa: I can not see the hook being used under wizard folder, is there a missing validation? if is not needed you can remove it from the test.

src/authz-module/wizard/AssignRoleWizard.test.tsx
```diff
@@ -0,0 +1,196 @@
...
+    mockUseValidateUserPermissions.mockReturnValue({
+      data: permissionsFor({ library: true }),
+    });
+  });
+
+  it('Cancel returns to the previous view', async () => {
+    const user = userEvent.setup();
+    const onClose = jest.fn();
+    renderWizard({ onClose });
+    await user.click(screen.getAllByRole('button', { name: /Cancel/i })[0]);
```
dcoa:
```suggestion
    await user.click(screen.getByRole('button', { name: /cancel/i }));
```

src/authz-module/role-assignation-wizard/AssignRoleWizard.test.tsx
```diff
@@ -0,0 +1,196 @@
...
+    mockValidateMutateAsync.mockResolvedValue({ invalidUsers: [], validUsers: ['alice'] });
+    renderWizard();
+    await user.type(screen.getByTestId('users-input'), 'alice');
+    await user.click(screen.getByTestId('role-option-library_admin'));
+    await user.click(screen.getByTestId('role-option-course_admin'));
+    await user.click(screen.getByRole('button', { name: /Next/i }));
+    await waitFor(() => {
+      expect(screen.getByTestId('stepper')).toHaveAttribute('data-active-step', 'define-application-scope');
+    });
+  });
```
dcoa, line 58: You can create some utilities

```suggestion
// selectors
const getUsersInput = () => screen.getByLabelText(/Add users by username or email/i);
const getRoleRadio = (name: RegExp) => screen.getByRole('radio', { name });
const getNextButton = () => screen.getByRole('button', { name: /^Next$/i });

 it('selecting a different role replaces the previous selection', async () => {
    const user = userEvent.setup();
    mockValidateMutateAsync.mockResolvedValue({ invalidUsers: [], validUsers: ['alice'] });
    renderWizard();
    await user.type(getUsersInput(), 'alice');
    await user.click(getRoleRadio(/Library Admin/i));
    await user.click(getRoleRadio(/Course Admin/i));
    expect(getRoleRadio(/Course Admin/i)).toBeChecked();
    expect(getRoleRadio(/Library Admin/i)).not.toBeChecked();
  });
```

Same for the other test cases

src/authz-module/wizard/AssignRoleWizardPage.test.tsx
```diff
@@ -0,0 +1,87 @@
...
+  ...jest.requireActual('react-router-dom'),
+  useSearchParams: jest.fn(),
+  useNavigate: jest.fn(),
+}));
+
+jest.mock('../data/hooks', () => ({
+  useLibrary: jest.fn(),
+}));
+
+jest.mock('./AssignRoleWizard', () => ({
```
dcoa: Avoid mocking components.

src/authz-module/wizard/AssignRoleWizardPage.test.tsx
```diff
@@ -0,0 +1,87 @@
...
+    const mockNavigate = jest.fn();
+    const { useNavigate } = jest.requireMock('react-router-dom');
+    useNavigate.mockReturnValue(mockNavigate);
+
+    const user = userEvent.setup();
+    renderWrapper(<AssignRoleWizardPage />);
+    await user.click(screen.getByTestId('wizard-close'));
+
+    expect(mockNavigate).toHaveBeenCalledWith('/authz');
+  });
```
dcoa:
```suggestion
  it('navigates to home path when the wizard Cancel is clicked', async () => {
    const { navigate } = setupMocks();
    const user = userEvent.setup();
    renderWrapper(<AssignRoleWizardPage />);
    await user.click(screen.getByRole('button', { name: /Cancel/i }));
    expect(navigate).toHaveBeenCalledWith('/authz');
  });
```

src/authz-module/wizard/AssignRoleWizardPage.test.tsx
```diff
@@ -0,0 +1,87 @@
...
+  users = '',
+  library = {
+    id: 'lib:123', title: 'Test Library', org: 'org', slug: 'test-lib',
+  },
+} = {}) => {
+  const { useSearchParams, useNavigate } = jest.requireMock('react-router-dom');
+  useSearchParams.mockReturnValue([new URLSearchParams(`scope=${scope}&users=${users}`)]);
+  useNavigate.mockReturnValue(jest.fn());
+  mockUseLibrary.mockReturnValue({ data: library });
+};
```
dcoa:
```suggestion
const setupMocks = ({ users = '' } = {}) => {
  const { useSearchParams, useNavigate } = jest.requireMock('react-router-dom');
  useSearchParams.mockReturnValue([new URLSearchParams(`users=${users}`)]);
  const navigate = jest.fn();
  useNavigate.mockReturnValue(navigate);
  return { navigate };
};
```

src/authz-module/wizard/AssignRoleWizardPage.test.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+describe('AssignRoleWizardPage', () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+
+  it('renders the wizard when library is present', () => {
+    setupMocks();
+    renderWrapper(<AssignRoleWizardPage />);
+    expect(screen.getByTestId('assign-role-wizard')).toBeInTheDocument();
```
dcoa:
```suggestion
    expect(screen.getByRole('heading', { name: 'Assign Role' })).toBeInTheDocument();
    expect(screen.getByLabelText(/Add users by username or email/i)).toBeInTheDocument();
```

src/authz-module/wizard/messages.ts
```diff
@@ -0,0 +1,139 @@
...
+    description: 'Back button label in the wizard',
+  },
+  'wizard.button.next': {
+    id: 'wizard.button.next',
+    defaultMessage: 'Next',
+    description: 'Next button label in the wizard',
+  },
+  'wizard.button.next.pending': {
+    id: 'wizard.button.next.pending',
+    defaultMessage: 'Validating...',
```
dcoa:
```suggestion
    defaultMessage: 'Validating',
```

src/authz-module/wizard/messages.ts
```diff
@@ -0,0 +1,139 @@
...
+    description: 'Next button label while user validation is in progress',
+  },
+  'wizard.button.save': {
+    id: 'wizard.button.save',
+    defaultMessage: 'Save',
+    description: 'Save button label in the wizard',
+  },
+  'wizard.button.save.pending': {
+    id: 'wizard.button.save.pending',
+    defaultMessage: 'Saving...',
```
dcoa:
```suggestion
    defaultMessage: 'Saving',
```

src/authz-module/wizard/SelectUsersAndRoleStep.tsx
```diff
@@ -0,0 +1,145 @@
...
+      <Form.Group controlId="users-input">
+        <Form.Label>{intl.formatMessage(messages['wizard.step1.users.label'])}</Form.Label>
+        <HighlightedUsersInput
+          ref={inputRef}
+          id="users-input"
+          value={users}
+          onChange={setUsers}
+          invalidUsers={invalidUsers}
+          placeholder={intl.formatMessage(messages['wizard.step1.users.placeholder'])}
+          hasNetworkError={!!validationError || invalidUsers.length > 0}
```
dcoa: I'm not a big fan of this presentation of the error in terms of the service availability (network errors), because can be confused as a validation error.

Image

I think is better having a separate component for it, maybe a  alert on top or maybe the toast message (if still relevant for the redesign)

I would like to have others option including @maguilaruxui

src/authz-module/constants.test.ts
```diff
@@ -0,0 +1,17 @@
+import { courseRolesMetadata } from './constants';
```
dcoa: I don't think we need to test `constants`

src/authz-module/roles-permissions/library/constants.ts
```diff
@@ -26,10 +26,18 @@ export const CONTENT_LIBRARY_PERMISSIONS = {
 // Note: this information will eventually come from the backend API
 // but for the MVP we decided to manage it in the frontend
 export const libraryRolesMetadata: RoleMetadata[] = [
-  { role: 'library_admin', name: 'Library Admin', description: 'The Library Admin has full control over the library, including managing users, modifying content, and handling publishing workflows. They ensure content is properly maintained and accessible as needed.' },
-  { role: 'library_author', name: 'Library Author', description: 'The Library Author is responsible for creating, editing, and publishing content within a library. They can manage tags and collections but cannot delete libraries or manage users.' },
-  { role: 'library_contributor', name: 'Library Contributor', description: 'The Library Contributor can create and edit content within a library but cannot publish it. They support the authoring process while leaving final publishing to Authors or Admins.' },
-  { role: 'library_user', name: 'Library User', description: 'The Library User can view and reuse content but cannot edit or delete any resource.' },
+  {
```
dcoa, line 29: For now is okay, but we need to consider move this file to a upper scope.

.eslintrc.js
```diff
@@ -1,4 +1,8 @@
 // eslint-disable-next-line import/no-extraneous-dependencies
 const { createConfig } = require('@openedx/frontend-build');
 
-module.exports = createConfig('eslint');
+module.exports = createConfig('eslint', {
```
dcoa: This change does not seems as part of the scope of the PR, could you explain why this change is important?

dcoa (changes requested):

src/authz-module/roles-permissions/library/constants.ts
```diff
@@ -26,10 +26,18 @@ export const CONTENT_LIBRARY_PERMISSIONS = {
 // Note: this information will eventually come from the backend API
```
dcoa, line 26: There is another thing here:

- `libraryRolesMetadata` lives under `authz-module/roles-permissions/libraries` folder 
- `courseRolesMetadata` lives under `authz-module` folder

If we want to keep consistency  `authz-module/roles-permissions/courses` 


Also I suggest use single instead of plural.

src/authz-module/wizard/HighlightedUsersInput.tsx
dcoa: @bra-i-am, sorry if I was not clear but I was talking about the whole file not only this line, avoid using inline styles, at least you really need to change something at runtime.  This `INPUT_STYLE` can perfectly be a custom class, there is things can can be manage with Paragon too, for example:

```tsx
 <div style={{ position: 'relative' }}>
```
Can be changed to https://paragon-openedx.netlify.app/foundations/layout/#position 

```tsx
 <div className="position-relative">
```

src/authz-module/wizard/SelectUsersAndRoleStep.tsx
maguilarUXUI: Error (2)
When is happened this kind of error? In the proposal the error happen when the user click next and the user is not part of the initiative

dcoa: Yes, and the validation error is presented as is showing in the proposal design. I'm not discussing that.

What I am talking about is error `500`, the validation service is not available.  I think those ones should be presented in another way, such as Alert on top instead of the red border and red message at the bottom.

@maguilarUXUI

arbrandes (changes requested):
Overall the implementation is solid. But I found a few issues with Claude's help (see below).  Also, can you please add some explicit test instructions so I know what to look for?

## Acceptance criteria gaps

The PR resolves most of the Step 1 acceptance criteria, but three items from [issue #91](https://github.com/openedx/frontend-app-admin-console/issues/91) are not met:

### 1. Pre-populate Step 1 when arriving from the user audit view

AC: "When accessing the wizard from the user audit view, the user input in Step 1 is pre-populated with that user."

[`AssignRoleWizardPage.tsx#L11-L12`](https://github.com/openedx/frontend-app-admin-console/blob/383375bfba1b71199a8c99694819f6c5db072c33/src/authz-module/wizard/AssignRoleWizardPage.tsx#L11-L12) reads `?users=` off the URL and forwards it to the wizard, but nothing in the diff actually navigates to the wizard with that query param. `git diff master -- src/authz-module/libraries-manager/` is empty, so `LibrariesUserManager` still cannot launch the wizard with a pre-filled user. The plumbing exists; the caller does not.

### 2. Role groups filtered by the current user's scopes

AC: "A user only sees the groups they have permissions to assign. If they have no library scopes, they do not see libraries. If they have no course scopes, they do not see courses."

[`AssignRoleWizard.tsx#L16`](https://github.com/openedx/frontend-app-admin-console/blob/383375bfba1b71199a8c99694819f6c5db072c33/src/authz-module/wizard/AssignRoleWizard.tsx#L16) passes the unconditional union `[...courseRolesMetadata, ...libraryRolesMetadata]` to the step, and no permission lookup is performed anywhere in the wizard. The two ❌ rows in the PR body's test table (`shows only course roles when user only has manage_course_team permission` / `shows only library roles when user only has manage_library_team permission`) confirm this is a known gap.

### 3. Breadcrumb and Cancel must return to the previous view

AC: "The wizard has a breadcrumb that exits the flow and returns to the previous view at any point." / "A Cancel button is available at the bottom of both steps. Clicking it returns to the previous view."

Both actions hardcode `ROUTES.HOME_PATH`:

- Breadcrumb: [`AssignRoleWizardPage.tsx#L17`](https://github.com/openedx/frontend-app-admin-console/blob/383375bfba1b71199a8c99694819f6c5db072c33/src/authz-module/wizard/AssignRoleWizardPage.tsx#L17)
- Cancel: [`AssignRoleWizardPage.tsx#L24`](https://github.com/openedx/frontend-app-admin-console/blob/383375bfba1b71199a8c99694819f6c5db072c33/src/authz-module/wizard/AssignRoleWizardPage.tsx#L24)

A user who reaches the wizard from `LibrariesUserManager` or `LibrariesTeamManager` will be dropped on `/authz` rather than returned to the scoped view they came from. Consider `navigate(-1)` or threading the origin through the wizard URL (e.g., a `from=` query param) so the exit paths honor the AC.

## Issues

### 1. Drop all `console.log` statements before landing

Two debug logs remain in the diff and should be removed:

- [`DefineApplicationScopeStep.tsx#L12`](https://github.com/openedx/frontend-app-admin-console/blob/383375bfba1b71199a8c99694819f6c5db072c33/src/authz-module/wizard/DefineApplicationScopeStep.tsx#L12) — `console.log(selectedRole, selectedScopes, onScopeToggle);` runs on every render once the user reaches Step 2.
- [`AssignRoleWizard.tsx#L97-L99`](https://github.com/openedx/frontend-app-admin-console/blob/383375bfba1b71199a8c99694819f6c5db072c33/src/authz-module/wizard/AssignRoleWizard.tsx#L97-L99) — `console.log('[AssignRoleWizard] handleSave', ...)` is guarded by an `eslint-disable-next-line` and should be replaced with a `TODO` for the follow-up API call.

### 2. Network `validationError` is not cleared when the user edits the input

[`src/authz-module/wizard/AssignRoleWizard.tsx#L45-L48`](https://github.com/openedx/frontend-app-admin-console/blob/383375bfba1b71199a8c99694819f6c5db072c33/src/authz-module/wizard/AssignRoleWizard.tsx#L45-L48)

```tsx
const handleUsersChange = useCallback((value: string) => {
  setInvalidUsers((prev) => (prev.length > 0 ? [] : prev));
  setUsers(value);
}, []);
```

`invalidUsers` is cleared on edit (covered by the `clears the invalid user highlights when the input is edited` test), but `validationError` is not. After a transient network failure the red "An error occurred while validating users..." line remains visible while the user types a new address, which is inconsistent with how invalid-user feedback behaves and can be confusing. Reset `validationError` here as well.

### 3. Catch-all `*` route is declared before `ASSIGN_ROLE_WIZARD_PATH`

[`src/authz-module/index.tsx#L33-L34`](https://github.com/openedx/frontend-app-admin-console/blob/383375bfba1b71199a8c99694819f6c5db072c33/src/authz-module/index.tsx#L33-L34)

```tsx
<Route path="*" element={<NotFoundError />} />
<Route path={ROUTES.ASSIGN_ROLE_WIZARD_PATH} element={<AssignRoleWizardPage />} />
```

React Router v6 ranks matches by specificity regardless of source order, so the wizard route does win and this works today. But declaring `"*"` in the middle of the list is a footgun: a future edit that (for example) replaces `Routes` with source-order matching, or adds a child route that shares a prefix, will silently break. Move the catch-all to be the last `<Route>` in the block.

### 4. Dead fallback for `contextType`

[`src/authz-module/wizard/SelectUsersAndRoleStep.tsx#L48-L49`](https://github.com/openedx/frontend-app-admin-console/blob/383375bfba1b71199a8c99694819f6c5db072c33/src/authz-module/wizard/SelectUsersAndRoleStep.tsx#L48-L49)

```tsx
const rolesByContext = roles.reduce<Record<string, RoleMetadata[]>>((acc, role) => {
  const context = role.contextType || 'library';
```

Every entry in `courseRolesMetadata` and `libraryRolesMetadata` now sets `contextType` explicitly, and the only caller passes those arrays, so the `|| 'library'` fallback is unreachable and misleading (a future course role added without `contextType` would silently render under Libraries). Either drop the fallback or tighten `RoleMetadata.contextType` to a required field.

src/authz-module/roles-permissions/library/constants.ts
bra-i-am: @dcoa, you're right; here are the changes applied https://github.com/openedx/frontend-app-admin-console/pull/109/changes/d9b405b0893f006b06edfb55efd140174b0852ac

bra-i-am:
@dcoa and @arbrandes, thank you so much for your time reviewing this PR :sparkles:

@dcoa: I already addressed the changes to keep consistency with `courseRolesMetadata`  https://github.com/openedx/frontend-app-admin-console/pull/109/changes/d9b405b0893f006b06edfb55efd140174b0852ac

@arbrandes: I already addressed the changes you asked for https://github.com/openedx/frontend-app-admin-console/pull/109/changes/f6b054f8f6bffd0271749c8dac3ea342ed2ac567. Here is a summary of all the points with a brief description of each one

| # | Item | Status |
|---|------|--------|
| AC-1 | Pre-populate Step 1 when arriving from user audit view | ✅ Fixed — `LibrariesUserManager` now navigates to `buildWizardPath({ users, from })`, pre-filling the username and setting the return path |
| AC-2 | Role groups filtered by the current user's scopes | ⚠️ Partial — `AssignRoleWizard` now accepts a `roles` prop so callers can pass a filtered subset; full filtering is blocked on a permission-lookup API that doesn't exist yet |
| AC-3 | Breadcrumb and Cancel must return to the previous view | ✅ Fixed — `AssignRoleWizardPage` reads `?from=` and uses it for both the breadcrumb `to` and the `onClose` handler; defaults to `ROUTES.HOME_PATH` when absent |
| I-1a | `console.log` in `DefineApplicationScopeStep.tsx` | ✅ Fixed — removed; replaced with a `TODO` comment |
| I-1b | `console.log` in `AssignRoleWizard.tsx` `handleSave` | ✅ Fixed — removed along with the `eslint-disable-next-line`; `TODO` comment now names the three values needed for the API call |
| I-2 | Network `validationError` not cleared when input is edited | ✅ Fixed — `handleUsersChange` now calls `setValidationError(null)`; covered by a new test |
| I-3 | Catch-all `*` route declared before `ASSIGN_ROLE_WIZARD_PATH` | ✅ Fixed — `*` moved to be the last `<Route>` in the block |
| I-4 | Dead `\|\| 'library'` fallback for `contextType` | ✅ Fixed — `contextType` made required in `RoleMetadata` (in `types.ts` and the local interface), `contextType` added to all `rolesObject`/`rolesLibraryObject` entries, fallback removed |

src/authz-module/wizard/SelectUsersAndRoleStep.tsx
maguilarUXUI: In this case I think the best option is a toast with the retry option if is possible.

bra-i-am: thanks @dcoa and @maguilarUXUI :sparkles:

I already added the toast:

image

src/authz-module/wizard/AssignRoleWizard.tsx
```diff
@@ -0,0 +1,196 @@
...
+  const validateUsersMutation = useValidateUsers();
+
+  const handleUsersChange = useCallback((value: string) => {
+    setInvalidUsers((prev) => (prev.length > 0 ? [] : prev));
+    setValidationError(null);
+    setUsers(value);
+  }, []);
+
+  const handleClose = () => {
+    setActiveStep(STEPS.SELECT_USERS_AND_ROLE);
```
jesusbalderramawgu: I suggest here to have an initial state with the default values and just call that inside this function, I think the way it is right now is more brittle.
```

const getInitialState = () => ({
  activeStep: STEPS.SELECT_USERS_AND_ROLE,
  users: initialUsers,
  selectedRole: null,
  selectedScopes: new Set<string>(),
  validationError: null,
  invalidUsers: [],
  validatedUsers: [],
});
```

src/authz-module/role-assignation-wizard/AssignRoleWizard.tsx
```diff
@@ -0,0 +1,196 @@
...
+      const result = await validateUsersMutation.mutateAsync({ data: { users: usersList } });
+      if (result.invalidUsers?.length > 0) {
+        setInvalidUsers(result.invalidUsers);
+      } else {
+        setValidatedUsers(usersList);
+        setActiveStep(STEPS.DEFINE_APPLICATION_SCOPE);
+      }
+    } catch (error) {
+      setValidationError(intl.formatMessage(messages['wizard.validate.error']));
+      showErrorToast(error, validateUsersAndProceed);
```
jesusbalderramawgu, line 98: is this correct? here you are passing the same function as a retry handler

src/authz-module/role-assignation-wizard/AssignRoleWizard.tsx
```diff
@@ -0,0 +1,196 @@
...
+  const [validatedUsers, setValidatedUsers] = useState<string[]>([]);
+
+  const usersInputRef = useRef<HTMLTextAreaElement>(null);
+
+  const validateUsersMutation = useValidateUsers();
+
+  const handleUsersChange = useCallback((value: string) => {
+    setInvalidUsers((prev) => (prev.length > 0 ? [] : prev));
+    setValidationError(null);
+    setUsers(value);
```
jesusbalderramawgu, line 68: not sure but I think you should add here a `setValidatedUsers([])`
because validateUsers becomes stale if user edits the input after validation

src/authz-module/wizard/AssignRoleWizard.tsx
```diff
@@ -0,0 +1,196 @@
...
+  // TODO: replace with real assignment API call using validatedUsers, selectedRole, selectedScopes
+  const handleSave = () => {
+    if (!selectedRole || selectedScopes.size === 0 || validatedUsers.length === 0) { return; }
+    handleClose();
+  };
+
+  const isError = invalidUsers.length > 0 || !!validationError;
+
+  useEffect(() => {
+    if (isError && usersInputRef.current) {
```
jesusbalderramawgu: this will trigger on any error change even unrelates one, don't you need to refine it a little bit?
something like
`if ((invalidUsers.length > 0 || validationError) && usersInputRef.current)`

src/authz-module/wizard/AssignRoleWizard.tsx
```diff
@@ -0,0 +1,196 @@
...
+      usersInputRef.current.scrollIntoView({ behavior: 'smooth', block: 'center' });
+    }
+  }, [isError]);
+
+  const canProceed = !!users.trim() && !!selectedRole;
```
jesusbalderramawgu: here if we have a string "a,,," it passes right?
maybe you could do something like
```
const parsed = parseUsers(users);
const canProceed = parsed.length > 0 && !!selectedRole;
```

src/authz-module/wizard/AssignRoleWizard.tsx
```diff
@@ -0,0 +1,196 @@
...
+    setUsers('');
+    setSelectedRole(null);
+    setSelectedScopes(new Set());
+    setValidationError(null);
+    setInvalidUsers([]);
+    setValidatedUsers([]);
+    onClose();
+  };
+
+  const parseUsers = (input: string): string[] => input
```
jesusbalderramawgu: this is recreated on every render, I don't see any dependency so you could jus hoist it outside

src/authz-module/role-assignation-wizard/components/HighlightedUsersInput.tsx
```diff
@@ -0,0 +1,116 @@
...
+            pointerEvents: 'none',
+            zIndex: 0,
+          }}
+        >
+          {renderedParts}
+        </div>
+      )}
+
+      {/* Actual textarea — text is transparent when overlay is active */}
+      <textarea
```
jesusbalderramawgu, line 67: missing ARIA attributes

jesusbalderramawgu, line 67: also, wouldn't be better to use the textarea from paragon? instead of this?

src/authz-module/wizard/HighlightedUsersInput.tsx
```diff
@@ -0,0 +1,116 @@
+import {
+  useMemo, useRef, useCallback, forwardRef,
+} from 'react';
+
+// Shared styles between overlay and textarea to keep text positions in sync
+const INPUT_STYLE: React.CSSProperties = {
```
jesusbalderramawgu: for styling you should add paragon classes or if you need custom css, you should add it to the scss files

src/authz-module/wizard/HighlightedUsersInput.tsx
```diff
@@ -0,0 +1,116 @@
...
+      {/* Highlight layer — sits behind the transparent textarea */}
+      {hasHighlights && (
+        <div
+          ref={overlayRef}
+          aria-hidden="true"
+          style={{
+            ...INPUT_STYLE,
+            position: 'absolute',
+            inset: 0,
+            background: '#fff',
```
jesusbalderramawgu: please check my comment above about styles

src/authz-module/wizard/SelectUsersAndRoleStep.tsx
```diff
@@ -0,0 +1,133 @@
...
+  inputRef,
+}: SelectUsersAndRoleStepProps) => {
+  const intl = useIntl();
+
+  const contextLabels: Record<string, string> = {
+    library: intl.formatMessage(messages['wizard.step1.roles.contextLabel.library']),
+    course: intl.formatMessage(messages['wizard.step1.roles.contextLabel.course']),
+  };
+
+  const rolesByContext = roles.reduce<Record<string, RoleMetadata[]>>((acc, role) => {
```
jesusbalderramawgu: this has a performance issue, it runs on every render.
you could wrap it in useMemo using roles as dependency

src/authz-module/wizard/SelectUsersAndRoleStep.tsx
```diff
@@ -0,0 +1,133 @@
...
+          <h5 className="h4 text-gray role-group-header my-4 pl-4">
+            {contextLabels[context] || context}
+          </h5>
+          <Form.RadioSet
+            name={`role-selection-${context}`}
+            value={selectedRole || ''}
+            onChange={(e) => setSelectedRole(e.target.value)}
+            className="pl-4"
+          >
+            {rolesByContext[context].map((role) => {
```
jesusbalderramawgu: I suggest this to be safer
`(rolesByContext[context] || []).map()`

src/authz-module/wizard/SelectUsersAndRoleStep.tsx
```diff
@@ -0,0 +1,133 @@
...
+  const rolesByContext = roles.reduce<Record<string, RoleMetadata[]>>((acc, role) => {
+    const context = role.contextType;
+    if (!acc[context]) { acc[context] = []; }
+    acc[context].push(role);
+    return acc;
+  }, {});
+
+  const orderedContexts = CONTEXT_ORDER.filter((ctx) => rolesByContext[ctx]);
+
+  const adminUrl = `${getConfig().LMS_BASE_URL}/admin`;
```
jesusbalderramawgu: you could move this outside the component

src/authz-module/role-assignation-wizard/components/SelectUsersAndRoleStep.tsx
```diff
@@ -0,0 +1,133 @@
...
+
+  const orderedContexts = CONTEXT_ORDER.filter((ctx) => rolesByContext[ctx]);
+
+  const adminUrl = `${getConfig().LMS_BASE_URL}/admin`;
+
+  return (
+    <div className="select-users-and-role-step">
+      {/* Users Section */}
+      <h3 className="text-primary mb-2">{intl.formatMessage(messages['wizard.step1.users.heading'])}</h3>
+      <Form.Group controlId="users-input">
```
jesusbalderramawgu, line 54: the form is missing ARIA attributes

rodmgwgu (commented):
Tested in my local and functionality looks good, I'll approve once Jesus' comments are addressed. Thanks!

src/authz-module/libraries-manager/messages.ts
```diff
@@ -106,6 +106,11 @@ const messages = defineMessages({
     defaultMessage: 'Retry',
     description: 'Label for retry button.',
   },
+  'library.authz.manage.assign.role.wizard.button': {
```
jacobo-dominguez-wgu: I do not see where this message is being used.

src/authz-module/role-assignation-wizard/AssignRoleWizard.tsx
```diff
@@ -0,0 +1,196 @@
...
+
+  const handleScopeToggle = useCallback((scopeId: string) => {
+    setSelectedScopes((prev) => {
+      const next = new Set(prev);
+      if (next.has(scopeId)) { next.delete(scopeId); } else { next.add(scopeId); }
+      return next;
+    });
+  }, []);
+
+  // TODO: replace with real assignment API call using validatedUsers, selectedRole, selectedScopes
```
jacobo-dominguez-wgu, line 110: Is this TODO still pending?

jacobo-dominguez-wgu:
In the figma design, when there is an error, an error icon is shown on the step number, I do not see it implemented.

Figma design:
image


Actual UI: 
image

src/authz-module/role-assignation-wizard/AssignRoleWizard.test.tsx
jacobo-dominguez-wgu, line 1: I would suggest a more descriptive name for this folder, like `role-assignation-wizard` or `assign-role-wizard`, `role-assignation`.

jacobo-dominguez-wgu: And try to keep the main component files on the root of this folder and for the rest of the components create a `/components` folder

src/authz-module/role-assignation-wizard/AssignRoleWizard.tsx
bra-i-am: @jesusbalderramawgu, yees, that's the expected behavior when there is an unexpected error (e.g. 5xx)

https://github.com/openedx/frontend-app-admin-console/pull/109#discussion_r3102192311

src/authz-module/wizard/AssignRoleWizardPage.tsx
```diff
@@ -0,0 +1,35 @@
...
+import { ROUTES } from '../constants';
+import messages from './messages';
+
+const AssignRoleWizardPage = () => {
+  const intl = useIntl();
+  const navigate = useNavigate();
+  const [searchParams] = useSearchParams();
+  const initialUsers = searchParams.get('users') || '';
+  const raw = searchParams.get('from') ?? '';
+  const returnTo = raw.startsWith('/') ? raw : ROUTES.HOME_PATH;
```
rodmgwgu: Suggested by Claude: In AssignRoleWizardPage.tsx, there's a check raw.startsWith('/') to prevent external URLs. However, //evil.com also starts with / and would be treated as a protocol-relative URL if ever used in an <a href> or
  window.location. While navigate() from react-router likely handles this safely, the validation should be stricter:

  // Current
  const returnTo = raw.startsWith('/') ? raw : ROUTES.HOME_PATH;

  // Safer — reject protocol-relative URLs
  const returnTo = (raw.startsWith('/') && !raw.startsWith('//')) ? raw : ROUTES.HOME_PATH;

  The test covers https://evil.com but not //evil.com.

src/authz-module/role-assignation-wizard/AssignRoleWizard.tsx
```diff
@@ -0,0 +1,196 @@
...
+    setValidatedUsers([]);
+    onClose();
+  };
+
+  const parseUsers = (input: string): string[] => input
+    .split(',')
+    .map((u) => u.trim())
+    .filter(Boolean);
+
+  const validateUsersAndProceed = async () => {
```
rodmgwgu, line 82: Perhaps we should guard here and check for validateUsersMutation.isPending, to avoid double invocation which can happen when the user clicks quickly, which could happen even if we are disabling the button (there is a delay while React re-renders the button as disabled), or when this is called on the retry toast.

src/authz-module/roles-permissions/library/constants.ts
```diff
@@ -26,10 +26,18 @@ export const CONTENT_LIBRARY_PERMISSIONS = {
...
 // but for the MVP we decided to manage it in the frontend
 export const libraryRolesMetadata: RoleMetadata[] = [
```
rodmgwgu, line 28: This is also defined in src/authz-module/constants.ts, I think we should keep just one.

src/authz-module/role-assignation-wizard/AssignRoleWizard.tsx
bra-i-am: this is already covered; the only way to reach step 2 (where Save is available and a stale `validatedUsers` really matters) is through `validateUsersAndProceed`, which always overwrites validatedUsers with the fresh list. Do you think it is better to change it?

src/authz-module/constants.ts
```diff
@@ -140,3 +160,56 @@ export const DJANGO_MANAGED_ROLES = ['django.superuser', 'django.globalstaff'];
 export const TABLE_DEFAULT_PAGE_SIZE = 10;
 
 export const DEFAULT_FILTER_PAGE_SIZE = 5;
+// Course Permission Keys
+export const COURSE_PERMISSIONS = {
```
rodmgwgu: I don't see this being used in this PR, can we remove?

rodmgwgu (changes requested):

src/authz-module/role-assignation-wizard/components/HighlightedUsersInput.tsx
bra-i-am: not here... this component deliberately uses a raw <textarea> because it needs pixel-perfect alignment between the textarea and the overlay div for the individual usernames red highlighting. Paragon's Form.Control wraps the textarea in extra markup and applies its own styles, which would break that alignment :(

src/authz-module/role-assignation-wizard/AssignRoleWizard.tsx
bra-i-am: yep, this handleSave is addressed in the Step 2 PR

bra-i-am:
@rodmgwgu, @jesusbalderramawgu, @jacobo-dominguez-wgu thank you for your help reviewing this — I’ve addressed the feedback you shared. Please let me know if there’s anything else that should be resolved.

I’ll let the tests run, and I’ll take care of any issues tomorrow morning if anything comes up.

jacobo-dominguez-wgu (approved):
Thank you so much! I have tested it with manual valid and invalid usernames and emails input and with a preset user everything seems to be working well.

jacobo-dominguez-wgu:
Lets wait til tomorrow for @rodmgwgu´s comments before merging.

src/authz-module/role-assignation-wizard/AssignRoleWizard.tsx
```diff
@@ -0,0 +1,205 @@
...
+  selectedScopes: new Set<string>(),
+  invalidUsers: [] as string[],
+  validatedUsers: [] as string[],
+});
+
+const AssignRoleWizard = ({ onClose, initialUsers = '', roles = allRolesMetadata }: AssignRoleWizardProps) => {
+  const intl = useIntl();
+  const { showErrorToast } = useToastManager();
+  const [activeStep, setActiveStep] = useState<StepKey>(STEPS.SELECT_USERS_AND_ROLE);
+  const [users, setUsers] = useState(initialUsers);
```
jacobo-dominguez-wgu, line 55: Just a nit in this component, so many local states can be better handled with a `useReducer` or even use react context, but lets do that in an upcoming issue.

arbrandes (approved):
Thank for the test instructions, everything works exactly as described!

rodmgwgu (approved):
Thanks!

PR #140, opened 2026-04-23
fix: replaced space for + char to render the filter correctly in the UI

jesusbalderramawgu:
### Issue

image

### Description

The filter is showing spaces instead of "+" chars.
for example after the redirection to admin console this  string 
course-v1:TestOrg+101+2025_T1
is received as
course-v1:TestOrg 101 2025_T1

### Fix
- I made a change to keep the "+" char when the param is received

NOTE: after debugging I saw that we send the correct data

Screenshot 2026-04-23 at 3 56 05 p m
Backend print in AUTHZ view
Screenshot 2026-04-23 at 3 57 00 p m

UI
Screenshot 2026-04-23 at 4 05 30 p m

### DIFF
--- a/src/authz-module/authz-home/index.tsx
+++ b/src/authz-module/authz-home/index.tsx
@@ -12,7 +12,7 @@ const AuthzHome = () => {
   const { hash } = useLocation();
   const intl = useIntl();
   const [searchParams] = useSearchParams();
-  const presetScope = searchParams.get('scope') || undefined;
+  const presetScope = searchParams.get('scope')?.replace(/\s/g, '+') || undefined;
 
   const pageTitle = intl.formatMessage(messages['authz.manage.page.title']);
 
--- a/src/authz-module/authz-home/index.tsx
+++ b/src/authz-module/authz-home/index.tsx
@@ -12,6 +12,10 @@ const AuthzHome = () => {
   const { hash } = useLocation();
   const intl = useIntl();
   const [searchParams] = useSearchParams();
+
+  /* In the authoring repository the scope is encoded but this is to Replace spaces with '+'
+     to match URL encoding format in case the 'scope' parameter is provided with spaces instead of '+'
+  */
   const presetScope = searchParams.get('scope')?.replace(/\s/g, '+') || undefined;
 
   const pageTitle = intl.formatMessage(messages['authz.manage.page.title']);


openedx-webhooks:
Thanks for the pull request, @jesusbalderramawgu!


</details>

- Dependencies
- Blockers
- Timeline information
- Partner information
- Supporting documentation
</details>





</details><details>


---

</details>

- The need for product review

</details>

src/authz-module/authz-home/index.tsx
```diff
@@ -12,7 +12,7 @@ const AuthzHome = () => {
   const { hash } = useLocation();
   const intl = useIntl();
   const [searchParams] = useSearchParams();
-  const presetScope = searchParams.get('scope') || undefined;
+  const presetScope = searchParams.get('scope')?.replace(/\s/g, '+') || undefined;
```
rodmgwgu, line 19: This works here because course and library key parts (org, run, number), don't accept spaces or special characters. Otherwise, this would potentially break keys with spaces (which is not the case).

However the root cause of this issue is that we are not url-encoding the query params on the links generated in frontend-app-authoring.

If we did that, the course key would be encoded as "course-v1%3AOpenedX%2BDemoX%2BDemoCourse", and on decode, it will come back correctly here as + instead of spaces.

We can keep this change as a fallback, but I think we should fix the root cause in frontend-app-authoring

rodmgwgu (approved):
This works, but let's fix the root cause in frontend-app-authoring

src/authz-module/authz-home/index.tsx
jesusbalderramawgu: thanks @rodmgwgu , I checked first on authoring and I saw the URL correctly there, but let me double check.
agreed that we should use encodeURI on authoring repo.

jesusbalderramawgu: @rodmgwgu  you were right, I've added the change to this PR
https://github.com/openedx/frontend-app-authoring/pull/3023
and it works correct, thanks!

jacobo-dominguez-wgu:
Is this pr still needed after the encoding added in https://github.com/openedx/frontend-app-authoring/pull/3023?

jesusbalderramawgu:
> Is this pr still needed after the encoding added in [openedx/frontend-app-authoring#3023](https://github.com/openedx/frontend-app-authoring/pull/3023)?

I think we don't need it, but it Rodrigo mentioned that we could leave this as fallback.
I think we can leave it in case someone copies and pastes the url with "+" char
@rodmgwgu  what do you think?

rodmgwgu:
> > Is this pr still needed after the encoding added in [openedx/frontend-app-authoring#3023](https://github.com/openedx/frontend-app-authoring/pull/3023)?
> 
> I think we don't need it, but it Rodrigo mentioned that we could leave this as fallback. I think we can leave it in case someone copies and pastes the url with "+" char @rodmgwgu what do you think?

If we are keeping it, I think we should add a comment explaining it. Also I would like to hear @dcoa opinion on this.

dcoa:
I think having this as a fallback is good idea, usually the urls in learning or authoring has the `courseKey` as it is, not encoded, and users will feel more nature to copy and paste that. I agree to leave a explanation comment in the code. 

In the future we can have a scope resolver if we need to cover other cases besides the `'+' -> ' '`.

@rodmgwgu @jesusbalderramawgu @jacobo-dominguez-wgu

rodmgwgu (approved):
@jesusbalderramawgu Let's add the explanation comment and we are good to go, thanks!

jesusbalderramawgu:
Thank you, this has been updated

jacobo-dominguez-wgu (approved):

PR #166, opened 2026-06-24
chore: Configure Dependabot for GitHub Actions updates

bmtcril:
This will help keep Github actions up to date for security and prevent issues like Codecov changing something that breaks CI.


dcoa (approved):
Thank you for this

PR #172, opened 2026-06-29
fix: limit assignable roles to the user's manage-team permissions

dcoa:
# Description

The "Assign Role" entry point and the role list inside the assignment wizard were shown to every user regardless of whether they could actually manage a team. This PR gates both on the current user's permissions, so people only see the action and the role groups they're allowed to assign.

## Changes

**AddRoleButton**
- Validates `MANAGE_LIBRARY_TEAM` / `MANAGE_COURSE_TEAM` and renders the button only when at least one is allowed; otherwise renders nothing.

**AssignRoleWizardPage**
- Derives `rolesAssignable` from the permission response, library roles when library-team management is allowed, course roles when course-team management is allowed, and passes them to `AssignRoleWizard` via the existing roles prop.
- Replace the `allRolesMetadata` used as default value in roles props for an empty array. So in the meantime the validation is done the user don't see the complete list.

**RolesFilter**

- Validates `VIEW_LIBRARY_TEAM` / `VIEW_COURSE_TEAM` and renders the respecting roles according to the allowed scope.

### Context
It follows the proposal in https://github.com/openedx/wg-build-test-release/issues/603
It solves https://github.com/openedx/frontend-app-admin-console/issues/163
It solves https://github.com/openedx/frontend-app-admin-console/issues/161

### Testing instructions
1. As a non-staff user with a library_user role in a single scope, the Assign Roles button should not be visible, only library roles should be in the filters.
2. As a non-staff user with a library_admin role in a single scope, the Assign Roles button should be visible, and the Wizard/filters displays only library roles.
3. As a user with multi scope manage team permissions, such as supper admin should see the button and the roles for both scopes. 

>[!Important]
> Depends on https://github.com/openedx/openedx-authz/pull/348 - Available in version `openedx-authz==1.20.0`

PR #176, opened 2026-07-02
fix: hide course authoring content when enable_course_authoring is off

bra-i-am:
Closes https://github.com/openedx/openedx-authz/issues/340
Closes https://github.com/openedx/openedx-authz/issues/341

When `authz.enable_course_authoring` is off, the Admin Console was still displaying course-related content in filters, the Team Members table, the Audit User view, and the role assignment wizard. This PR makes all those surfaces aware of the flag state using the dedicated `GET /api/authz/v1/waffle-flag-states/` endpoint introduced in [openedx-authz#358](https://github.com/openedx/openedx-authz/pull/358).

## How it works

The flag state is fetched from a dedicated endpoint that returns the global value plus every org and course that has an active override, split into `on` and `off` lists:

```json
{
  "global": false,
  "org_overrides": { "on": ["Demo"], "off": [] },
  "course_overrides": { "on": [], "off": ["course-v1:testing+CT01+CT01-2024"] }
}
```

A new `useCourseAuthoringFlag` hook centralises all resolution logic and exposes three resolvers:

- `isCourseAuthoringEnabled` — true if the flag is on at any tier (global, org, or course). Used as a coarse domain-level gate for the role filter and assignment wizard.
- `isCourseEnabled(courseId)` — resolves enablement for a specific course using the precedence **course override → org override → global**. Used to filter individual courses out of the scope filter and wizard scope list.
- `isOrgAuthoringEnabled(org)` — true if the org is explicitly enabled, has at least one course with an explicit `on` override, or inherits a global `on` without an explicit org `off`. Used to filter the org filter for course-only users.

Flag-state resolution is intentionally separate from permission validation (`useViewTeamPermissions`), because Casbin state can lag behind the flag when automatic migration is disabled or when the global flag changes. See [ADR 0003](docs/decisions/0003-course-authoring-flag-enforcement.rst) for the full rationale.

## Main changes

- **`data/api.ts`** — adds `getCourseAuthoringFlagStates` and the `CourseAuthoringFlagStates` / `WaffleFlagOverrides` interfaces.
- **`data/hooks.ts`** — adds `useCourseAuthoringFlagStates`
- **`hooks/useCourseAuthoringFlag.ts`** — derives `isCourseAuthoringEnabled`, `isCourseEnabled`, `isOrgAuthoringEnabled`, `isLoading`, and `isError` from the flag states query. Surfaces a single error toast with retry on fetch failure.
- **`RolesFilter`** — course roles require both `isCourseViewAllowed` and `isCourseAuthoringEnabled`.
- **`ScopesFilter`** — filters each course individually with `isCourseEnabled(scope.externalKey)`; libraries are always included.
- **`OrgFilter`** — filters orgs with `isOrgAuthoringEnabled` for course-only users; users with library access always see all orgs.
- **`AssignRoleWizardPage`** — course roles gated by `isCourseAuthoringEnabled` (the flag) independently of management permissions.
- **`useScopeListData`** — filters the wizard's scope list client-side with `isCourseEnabled`
- **`ScopeList`** — empty state only shown once all pages have loaded, since client-side filtering can empty pages while later pages still contain visible scopes.
- **`docs/decisions/0003-course-authoring-flag-enforcement.rst`** — ADR documenting the enforcement approach, its temporary nature, rejected alternatives, and future removal criteria.

## Test plan

**Setup:** you need an instance with the `authz.enable_course_authoring` flag and at least one org or course override configured. The filter, scope, and wizard tests do not require migration — they read the flag state directly. The assignment list tests (TeamMembersTable, AuditUser) require migration to have run, either by enabling `ENABLE_AUTOMATIC_AUTHZ_COURSE_AUTHORING_MIGRATION = True` for org/course-level changes, or by running `authz_rollback_course_authoring` manually for global flag changes.

| Scenario | Expected behaviour |
|---|---|
| Flag OFF globally, no overrides | Role filter hides course groups; scope filter hides all courses; org filter hides all orgs for course-only users; wizard has no course roles |
| Flag OFF globally, org A forced ON | Courses in org A appear in scope/org filters and wizard; all other orgs/courses hidden |
| Flag ON globally, course X forced OFF | Course X hidden in scope filter and wizard scope list; all other courses visible |
| Flag ON globally, org B forced OFF | Org B and its courses hidden for course-only users |
| User has only library roles | Libraries appear in all filters regardless of flag state |
| User has both library and course roles, flag OFF | Library content visible everywhere; course content hidden |
| `waffle-flag-states` endpoint unreachable | Error toast appears with retry button; all course content hidden |
| Loading state (slow network) | Course content hidden until flag state resolves; library content unaffected |


openedx-webhooks:
Thanks for the pull request, @bra-i-am!


</details>

- Dependencies
- Blockers
- Timeline information
- Partner information
- Supporting documentation
</details>





</details><details>


---

</details>

- The need for product review

</details>

src/authz-module/team-members/TeamMembersTable.test.tsx
```diff
@@ -194,6 +207,28 @@ describe('TeamMembersTable', () => {
     expect(mockNavigate).toHaveBeenCalledWith('/authz/user/johndoe');
   });
 
+  it('renders safely when role assignments data is undefined', () => {
+    // @ts-ignore
```
dcoa: Please avoid this

Instead:
```ts
import type { GetAllRoleAssignmentsResponse } from '@src/authz-module/data/api';

const mockedAllRoleAssignments: {
  data: GetAllRoleAssignmentsResponse | undefined;
  error: Error | null;
  isLoading: boolean;
  refetch: jest.Mock;
} = {
  // ...existing object
};
```

src/authz-module/hooks/useViewTeamPermissions.ts
```diff
@@ -0,0 +1,18 @@
...
+import {
+  CONTENT_COURSE_PERMISSIONS, CONTENT_LIBRARY_PERMISSIONS, VIEW_TEAM_PERMISSIONS,
+} from '@src/authz-module/roles-permissions';
+
+export const useViewTeamPermissions = () => {
+  const { data: permissions, isLoading } = useValidateUserPermissionsNonSuspense(VIEW_TEAM_PERMISSIONS);
+
+  const isCourseViewAllowed = permissions
+    ? permissions.some((p) => p.action === CONTENT_COURSE_PERMISSIONS.VIEW_COURSE_TEAM && p.allowed)
+    : true;
```
dcoa: The default value must be `false`. The reasoning is we can not assume a user has a permission.

It also make some parts of the UI for example the table flicks because `isCourseViewAllowed`  is `true` then evaluate to `false`.

src/authz-module/team-members/TeamMembersTable.tsx
```diff
@@ -43,12 +46,19 @@ const TeamMembersTable = ({ presetScope }: TeamMembersTableProps) => {
...
+  const { isCourseViewAllowed } = useViewTeamPermissions();
+
+  const effectiveQuerySettings = useMemo(() => {
+    if (isCourseViewAllowed || querySettings.roles) { return querySettings; }
+    return { ...querySettings, roles: LIBRARY_ROLE_KEYS };
+  }, [isCourseViewAllowed, querySettings]);
+
   const {
-    data: { results: roleAssignments, count } = { results: [], count: 0 },
+    data: { results: roleAssignments, count } = { results: [] as UserRole[], count: 0 },
```
dcoa: This type change is not necessary.

dcoa (changes requested):
Just  some small changes

rodmgwgu:
@bra-i-am does this change depend on new changes in openedx-authz? I tried to run it on my env which was built from main a couple of days ago and I see the following errors on the browser inspector when loading the admin-console form this branch:

Request: POST http://local.openedx.io:8000/api/authz/v1/permissions/validate/me

Payload: [{"action":"content_libraries.manage_library_team"},{"action":"courses.manage_course_team"}]

Response: [{"scope":["This field is required."]},{"scope":["This field is required."]}]

dcoa:
>  does this change depend on new changes in openedx-authz? 

@rodmgwgu yes, it depends on release 1.20.0 https://github.com/openedx/openedx-authz/releases/tag/v1.20.0

dcoa:
@MaferMazu @BryanttV    I have a question about the expected behavior. Specifically, is it okay to run the migrations when the flag is turned off?

I just want to make sure I understand all the requirements before reviewing.

I was able to see the assignment table filtered,  (although not the filter options when using the admin role) by enabling `ENABLE_AUTOMATIC_AUTHZ_COURSE_AUTHORING_MIGRATION` and testing the organization-level override.

image
image

However, I don't see the same behavior at the platform level. Am I missing a step, or is there something else that needs to be configured?

MaferMazu:
@dcoa, in my understanding, this solution works because there is no assignment of new authorization course authoring roles (thanks to the migration), so you won't be able to see the information. Is that so, @bra-i-am?

If so, I am a bit worried about it, because based on the comments in the issue, the criteria are not related to having a role; they are more related to the waffle flag.

> same fix as [#341](https://github.com/openedx/openedx-authz/issues/341) : we show Authoring UI if ANY Authoring related Waffle flag is ON. We hide it if ALL of them ar OFF.
> 
That's a quote from https://github.com/openedx/openedx-authz/issues/340#issuecomment-4754311973

It makes me think because you can definitely have the migration enabled, but if not, based on Guille's comment, you shouldn't be able to see anything related to new course authoring roles if the waffle flag is off.

For example, having the waffle flag active, assigning user Y as the course admin for course X, and then deactivating the waffle flag. - That user Y shouldn't be able to assign more users to the new authorization course authoring roles. <- This should be true regardless of whether migration conf is enabled.

I tested exactly that: I enabled the waffle flag globally, assigned another user as the course admin for a course, and then deactivated the waffle flag... and that "another user" is still able to assign course authoring roles. <- and this shouldn't be possible.

I remember a conversation where the proposal to solve this was to create a REST API in AuthZ that retrieves information about available domains (also taking into account the waffle flag). Also, we considered solving https://github.com/openedx/frontend-app-admin-console/pull/172 with that approach, but since it was resolved directly with front, I wonder if we can do something similar here, or if we would need the rest api to have the proper info from the back because the waffle flag info is trickier.

dcoa:
thanks @MaferMazu  for the extra context 

Mi opinion here:

> If so, I am a bit worried about it, because based on the comments in the issue, the criteria are not related to having a role; they are more related to the waffle flag.

The implementation in this PR is not bad understanding the automated migration run, and how it impact the backend. However is not enough covering all the test cases, based on what we are discussing. 


| Platform Flag | Org Override | Course Override | Effective Authoring State | User Has Permission? | Show Authoring Roles? |
| --- | --- | --- | --- | --- | --- |
| Off | None | None | Off | No | ❌ |
| Off | None | None | Off | Yes | ❌ |
| On | None | None | On | No | ❌ |
| On | None | None | On | Yes | ✅ |
| Off | Force On | None | On | No | ❌ |
| Off | Force On | None | On | Yes | ✅ |
| On | Force On | None | On | No | ❌ |
| On | Force On | None | On | Yes | ✅ |
| Off | Force Off | None | Off | No | ❌ |
| Off | Force Off | None | Off | Yes | ❌ |
| On | Force Off | None | Off | No | ❌ |
| On | Force Off | None | Off | Yes | ❌ |
| Off | None | Force On | On | No | ❌ |
| Off | None | Force On | On | Yes | ✅ |
| On | None | Force On | On | No | ❌ |
| On | None | Force On | On | Yes | ✅ |
| Off | None | Force Off | Off | No | ❌ |
| Off | None | Force Off | Off | Yes | ❌ |
| On | None | Force Off | Off | No | ❌ |
| On | None | Force Off | Off | Yes | ❌ |

| Platform Flag | Org Override | Course Override | Effective State | Reason |
| --- | --- | --- | --- | --- |
| On | Force Off | Force On | On | Course override has highest priority |
| On | Force On | Force Off | Off | Course override has highest priority |
| Off | Force On | Force Off | Off | Course override disables it |
| Off | Force Off | Force On | On | Course override enables it |
| On | Force Off | None | Off | Organization disables platform default |
| Off | Force On | None | On | Organization enables platform default |


> I remember a conversation where the proposal to solve this was to create a REST API in AuthZ that retrieves information about available domains (also taking into account the waffle flag). Also, we considered solving https://github.com/openedx/frontend-app-admin-console/pull/172 with that approach, but since it was resolved directly with front, I wonder if we can do something similar here, or if we would need the rest api to have the proper info from the back because the waffle flag info is trickier.

The other issue was a smaller scope that is why we can solved it,  even though we still needed  backend implementation.

The issue here is more complicated and I think need a solution at the backend side first, the reason is there is no platform standalone endpoint that we can use from the frontend to validate `waffle_flags` and cover all the test cases, for example there is `/api/toggles/v0/state/` but it needs Django staff role and there is not filters to help return only the required data.  (An endpoint like this could be useful in the MFE evolution, I imagine other modules evaluating other flags, but that is another discussion)


The platform has `/api/contentstore/v1/course_waffle_flags` it returns the platform level flags if I don't specify a course, but we can not distinguish between course level and org level, and we would then need to validate one by one of the scopes.


`assignments` endpoint already validates and return only library roles if the flag is _off_ , we can add a endpoint for `roles` that do the same but beyond that, same should happen with `scopes` and `organization` endpoints.


Scopes should be filtered properly, with that we can validate al front and there is not `course_key` then hide the course roles. (At the moment I did my test with **superuser**, the endpoint was returning the course for the organization and the course that has the flag _off_, even when `ENABLE_AUTOMATIC_AUTHZ_COURSE_AUTHORING_MIGRATION` setting is _on_)

Or even the `validate/me` me endpoint could check the flags before returning `is_allowed`? 
Let me know what do you think :)

MaferMazu:
@dcoa I agree.
We can prioritize modifying the validate/me endpoint, then the scope, and then the roles. So those endpoints should return information that takes into account the waffle flag.

mariajgrimaldi:
[This PR](https://github.com/openedx/openedx-authz/pull/361) was meant to partially solve [340](https://github.com/openedx/openedx-authz/issues/340) & [341](https://github.com/openedx/openedx-authz/issues/341) by supporting the proposal made here https://github.com/openedx/frontend-app-admin-console/pull/176#issuecomment-4900922914 for the Verawood release - returning False for scopes based on the waffle flag state. But the approach still needs to address other dimensions, such as the performance and correctness of the entire framework, especially since these issues are release blockers and the modified endpoints are critical to the system's functionality. Internally (thanks @dcoa!), we discussed the best solution moving forward, given that we shouldn't introduce big changes that put the framework's correctness at risk, and went ahead with #358 (thanks @MaferMazu!) and supporting filtering via the admin-console MFE. This will definitely introduce complexity to the MFE logic, which will be addressed in the next release cycle.

src/authz-module/components/TableControlBar/OrgFilter.tsx
```diff
@@ -11,15 +13,25 @@ const OrgFilter = ({
...
   } = useOrgs(searchValue, 1, DEFAULT_FILTER_PAGE_SIZE);
-  const filterChoices = useMemo(() => orgsData?.results?.map((org) => ({
-    displayName: org.name,
-    value: org.shortName,
-  })) || [], [orgsData]);
+
+  // Libraries span orgs and are always enabled, so they must keep their behavior: only
+  // filter orgs by the course-authoring flag for course-only users, and never while
+  // permissions or flag states are still loading (default to showing every org).
+  const filterByAuthoringFlag = !isLoading && !isFlagLoading && !isLibraryViewAllowed;
```
mariajgrimaldi, line 28: Does this work when a user has both library & course roles?

src/authz-module/components/TableControlBar/OrgFilter.tsx
```diff
@@ -11,15 +13,25 @@ const OrgFilter = ({
...
     data: orgsData = {
       count: 0, next: null, previous: null, results: [],
     },
   } = useOrgs(searchValue, 1, DEFAULT_FILTER_PAGE_SIZE);
-  const filterChoices = useMemo(() => orgsData?.results?.map((org) => ({
-    displayName: org.name,
-    value: org.shortName,
-  })) || [], [orgsData]);
+
+  // Libraries span orgs and are always enabled, so they must keep their behavior: only
```
mariajgrimaldi: Libraries span orgs and...?

src/authz-module/hooks/useCourseAuthoringFlag.ts
```diff
@@ -0,0 +1,85 @@
...
+    : false;
+
+  const { isCourseEnabled, isOrgAuthoringEnabled } = useMemo(() => {
+    const courseOn = new Set(flagStates?.courseOverrides.on ?? []);
+    const courseOff = new Set(flagStates?.courseOverrides.off ?? []);
+    const orgOn = new Set(flagStates?.orgOverrides.on ?? []);
+    const orgOff = new Set(flagStates?.orgOverrides.off ?? []);
+    const global = flagStates?.global ?? false;
+
+    const courseEnabled = (courseId: string): boolean => {
```
mariajgrimaldi, line 61: What about `courseFlagOverrideEnabled`?

src/authz-module/hooks/useCourseAuthoringFlag.ts
```diff
@@ -0,0 +1,85 @@
...
+      if (courseOff.has(courseId)) { return false; }
+      const org = orgOf(courseId);
+      if (org && orgOn.has(org)) { return true; }
+      if (org && orgOff.has(org)) { return false; }
+      return global;
+    };
+
+    const orgsWithForcedOnCourse = new Set([...courseOn].map(orgOf));
+
+    const orgAuthoringEnabled = (org: string): boolean => {
```
mariajgrimaldi, line 76: What about `orgFlagOverrideEnabled`?

src/authz-module/role-assignation-wizard/components/ScopeList.tsx
```diff
@@ -83,7 +83,9 @@ const ScopeList = ({
             />
           ))}
 
-          {orderedOrgs.length === 0 && (
+          {/* Loaded pages can be entirely filtered out client-side while later pages
+              still hold visible scopes, so only report empty once every page is in. */}
+          {orderedOrgs.length === 0 && !hasNextPage && !isFetchingNextPage && (
```
mariajgrimaldi, line 88: Does this affect the search? I mean, if I search for an org or course it's not on (forced off or globally off), would it appear in the results because it's on another page? 

This comment only makes sense when this is true: https://github.com/openedx/frontend-app-admin-console/pull/176/changes#r3595738419

mariajgrimaldi (commented):
Thank you so much for this! I'm currently rebuilding my environment to thoroughly test this, but in the meantime I left a few clarifying questions :)

src/authz-module/components/TableControlBar/OrgFilter.tsx
mariajgrimaldi: I'm testing and it works!

src/authz-module/hooks/useCourseAuthoringFlag.ts
```diff
@@ -0,0 +1,85 @@
...
+      || flagStates.orgOverrides.on.length > 0
+      || flagStates.courseOverrides.on.length > 0
+    : false;
+
+  const { isCourseEnabled, isOrgAuthoringEnabled } = useMemo(() => {
+    const courseOn = new Set(flagStates?.courseOverrides.on ?? []);
+    const courseOff = new Set(flagStates?.courseOverrides.off ?? []);
+    const orgOn = new Set(flagStates?.orgOverrides.on ?? []);
+    const orgOff = new Set(flagStates?.orgOverrides.off ?? []);
+    const global = flagStates?.global ?? false;
```
mariajgrimaldi, line 59: Did we also filter the role assignments list based on the flag as well? Does it make sense to implement it? I thought we were going to but it's not clear from the cover letter if that's our intent. 

Image

Although the scope `course-v1:Demo+DM+2026` doesn't appear in the filters or in the assignment wizard, it does in the assignment list, suggesting it might be on. One relevant point is that I can delete the assignment from the assignment list within the disabled scope, which is what we're trying to avoid: actions on disabled scopes.

mariajgrimaldi:
@bra-i-am: Do you think we can add test cases for each row in this table? https://github.com/openedx/frontend-app-admin-console/pull/176#issuecomment-4900922914

bra-i-am:
@mariajgrimaldi @dcoa @MaferMazu, I've been testing this using the table shared by Diana, and this is what I've found:

**Platform flag** [ CHECK :white_check_mark: ]
| Platform Flag | Org Override | Course Override | Effective Authoring State | User Has Permission? | Show Authoring Roles? | Evidence | Note |
| --- | --- | --- | --- | --- | --- | --- | --- |
| Off | None | None | Off | No | ❌ |image| Display library roles and scopes only. Table is filtered to show only libraries |
| Off | None | None | Off | Yes | ❌ |image| Display library roles and scopes only; the table displays the courses where the user has Casbin permissions |
| On | None | None | On | No | ❌ |image | Display library roles and scopes only |
| On | None | None | On | Yes | ✅ | image| Display all courses and libraries roles and scopes |

**ORG Override** [ CHECK :white_check_mark: ]
`TestOrg` is the one that will be overridden to test:

| Platform Flag | Org Override | Course Override | Effective Authoring State | User Has Permission? | Show Authoring Roles? | Evidence | Note |
| --- | --- | --- | --- | --- | --- | --- | --- |
| Off | Force On | None | On | No | ❌ |image| **admin:** can select in wizard only org courses with `Force on` **Regular user:** sees only library roles and scopes|
| Off | Force On | None | On | Yes | ✅ |image image| **admin:** can select in wizard only org courses with `Force on`. **Regular user:** can see scope and roles from the overridden orgs; the rows in the table are not filtered |
| On | Force On | None | On | No | ❌ | image| Display library roles and scopes only. Table is filtered to show only libraries|
| On | Force On | None | On | Yes | ✅ | image| Display all courses and libraries roles and scopes |
| Off | Force Off | None | Off | No | ❌ | image | Display library roles and scopes only |
| Off | Force Off | None | Off | Yes | ❌ | image | Displays only library roles and scopes despite the user having permissions in non-forced-off courses. The rows in the table are not filtered  |
| On | Force Off | None | Off | No | ❌ | image | Display library roles and scopes only |
| On | Force Off | None | Off | Yes | ❌ | image | Users can see the course roles and ONLY the ON scopes. The rows in the table are not filtered |

**Course Override** [ CHECK :white_check_mark: ]
| Platform Flag | Org Override | Course Override | Effective Authoring State | User Has Permission? | Show Authoring Roles? | Evidence | Note |
| --- | --- | --- | --- | --- | --- | --- | --- |
| Off | None | Force On | On | No | ❌ |image| Display library roles and scopes only |
| Off | None | Force On | On | Yes | ✅ | image image | **Admin:** can have access only to the `Force ON` courses. **Regular user:** can see all roles and ONLY the scopes with `Force ON` despite having permissions in more courses |
| On | None | Force On | On | No | ❌ |image| Display library roles and scopes only |
| On | None | Force On | On | Yes | ✅ | image | Can see all the roles and scopes |
| Off | None | Force Off | Off | No | ❌ |image| Display library roles and scopes only |
| Off | None | Force Off | Off | Yes | ❌ | image | Display library roles and scopes only. The rows in the table are not filtered |
| On | None | Force Off | Off | No | ❌ | image| Display library roles and scopes only |
| On | None | Force Off | Off | Yes | ❌ | image | User only has access to the courses that don't have `force OFF`. The rows in the table are not filtered |

---

**Override Priority (Cascade)**
| Platform Flag | Org Override | Course Override | Effective State | Reason | Evidence | Note |
| --- | --- | --- | --- | --- | --- | --- |
| On | Force Off | Force On | On | Course override has highest priority | image | It is only shown the course overridden with Force ON |
| On | Force On | Force Off | Off | Course override has highest priority | image | It is hidden the course overridden with Force OFF |
| On | Force Off | None | Off | Org Force Off overrides platform | image | |
| Off | Force On | Force Off | Off | Course override disables Org override, which overrides platform override | image | The image shows course roles because the org override is ON, but is not showing course scopes because the only course in that ON org is the one that is being Forced OFF |
| Off | Force Off | Force On | On | Course override enables the course | image | Course Force ON has priority over platform/org override; that's why the course is shown |
| Off | Force On | None | On | Organization enables platform default | image | Org override shows the courses of the Force ON org |

---

## Test Conclusions

### What works correctly

- **Scope filter** — per-course resolution (`isCourseEnabled`) is accurate in all tested scenarios. Force On shows only the enabled course; Force Off hides it while keeping other courses visible.
- **Org filter** — filters orgs correctly based on override state. Course-only users don't see disabled orgs.
- **Role filter + wizard** — correctly hides course roles and scopes when the flag is off at every tier (Platform=Off with no Force On anywhere).
- **Force On override at any tier** — restores course content correctly for users with permission, both in the wizard and filters.
- **Platform=On + Org/Course Force Off** — the roles filter correctly shows course roles because authoring is still enabled globally. The scope filter correctly hides the forced-off org's or course's scopes. The two gates work together: the domain gate answers "is authoring enabled anywhere?" (yes, Platform=On), and the per-scope gate answers "is this specific scope enabled?" (no, Force Off).
- **Override priority cascade** — course override always wins over org override, which always wins over platform default. Confirmed in all four cascade combinations.

### Confirmed known limitations

**Assignment list is not filtered** — in every scenario, the table shows Casbin rows regardless of the flag state. This is the documented limitation in ADR 0003 (Decision 4 + Implications section). Fixing it requires the backend to filter `/assignments/` by flag state, which is out of scope for this PR.

---

Overall, the behavior matches expectations across all tested scenarios. The only known limitation (assignment list not filtered by flag state) is documented in ADR 0003 and intentionally out of scope for this PR

src/authz-module/hooks/useCourseAuthoringFlag.ts
bra-i-am: @mariajgrimaldi I tried implementing client-side filtering on the assignment list, but it only introduces UX problems that are tricky to resolve cleanly: the table is server-driven and paginated, so filtering rows after they arrive from the API breaks the item count ("Showing 8 of 8" when only 6 are visible) and produces phantom pages (e.g. showing 2 pages when everything fits in 1, because the items all get filtered out client-side).

This is actually the reason ADR 0003 explicitly decided against it:

> The assignment lists are server-driven and paginated. Client-side filtering would invalidate pagination, counts, and sorting. The backend therefore remains responsible for hiding authoring-only assignments.

The intent is that Casbin migration (ADR 0013 in openedx-authz) keeps the assignment data aligned with the flag. Once migration runs, disabled-scope assignments won't be returned by the API at all, so the list stays clean without any frontend workaround.

That said, your concern about actions on disabled scopes is valid, so I've disabled the view action button for course assignments where `isCourseEnabled(scope)` returns `false`. This prevents navigating into a disabled scope to delete the assignment

The scenario you observed where the scope appears is still the known migration-lag window acknowledged in the ADR. The disabled button is the temporary guard I think we can apply without breaking the table UX, what do you say?

image

src/authz-module/role-assignation-wizard/components/ScopeList.tsx
bra-i-am: If you search for an org or course that is not ON, it won't appear because of [this line](https://github.com/openedx/frontend-app-admin-console/pull/176/changes#diff-06c06b79618c1ed37e863c5622d8fefc5a71a8611716436e1b9cb985dbf4e545R49)

src/authz-module/hooks/useCourseAuthoringFlag.ts
dcoa: To avoid the behavior described by Brayan and keep everything consistent the solution is to convert the tables to client-side, that is a bigger refactor. Furthermore I do not recommend.

I think that is a good solution @bra-i-am, ideally for both tables (global assignments and user specific assignments, we still can do deep navigation) and may be add a tooltip explained that assignment/actions over the course is disabled.

mariajgrimaldi: I think this makes sense. Considering we're not targeting changes in the backend, this works well!

mariajgrimaldi:
@bra-i-am: thanks a lot for all the testing! I don't want to block this further but I wonder whether we can implement all the manual testing using jest. In any case, this looks great to me! Thanks a lot for addressing all of my comments :)

bra-i-am:
@mariajgrimaldi, I added Jest coverage for those manual cases:

- Added a describe block inside `useCourseAuthoringFlag.test.tsx` covering every row of the truth table: platform flag on/off, org Force On/Off, course Force On/Off, and the cascade combinations
- Updated `ScopesFilter.test.tsx` where `isCourseEnabled` returns different values for different courses, asserting that the dropdown displays only the enabled

src/authz-module/hooks/useCourseAuthoringFlag.ts
dcoa: @bra-i-am @mariajgrimaldi I added the disabled rows in the user specific view and added a tooltip for explaining briefly why is disabled and what to do, for better usability.


**Before:**


[Screencast from 2026-07-20 16-17-55.webm](https://github.com/user-attachments/assets/eada2081-11d1-4e02-b28b-823aa3b8da6d)


**After**


[Screencast from 2026-07-20 16-22-58.webm](https://github.com/user-attachments/assets/833ad732-ef43-48ea-9194-3f8525e6ca12)


I am happy to merge if we all agree on the the implemented changes, specially the message

dcoa (approved):
The changes looks good to me, and I see all the concerns addressed and/or documented.

Pending decision https://github.com/openedx/frontend-app-admin-console/pull/176#discussion_r3612361613

src/authz-module/hooks/useCourseAuthoringFlag.ts
mariajgrimaldi: LGTM! Thank you so much for all the effort!

mariajgrimaldi (approved):
LGTM! Thanks for addressing all of my comments :)

mariajgrimaldi:
**EDIT:** I think it might be related to some cache cause it worked after some time!

I was re-testing locally and found that when the global flag is on I can't delete any of the roles. I first had the overrides on and it worked as expected, but then I turn on the global flag and it blocked all of the courses:

Screenshot from 2026-07-21 12-22-58
Screenshot from 2026-07-21 12-22-40
Screenshot from 2026-07-21 12-22-35

Issue #149, opened 2026-04-29
Convert Admin Console to frontend-base


arbrandes:
### Description

Convert the Admin Console app to use frontend-base, as per https://github.com/openedx/frontend-base/blob/main/docs/how_tos/migrate-frontend-app.md.

The converted code will land in master immediately once the PR is approved.

PR #210, opened 2026-08-19
Frontend base Migration

### LINKED ISSUE #211: Convert the app to frontend-base
### Description

The conversion proper: reworking the app's internal structure onto `site.config` and slots, porting whatever header and footer customizations it carries, and getting the test suite passing again against the new structure. The result lives on the long-lived `frontend-base` branch rather than on `master`.

This subsumes what #149 previously tracked as three separate steps, since the migration PR does all three at once.

The migration PR is #210.

jesusbalderramawgu:
### Description

PR to migrate this repository to frontend base. It closes this [issue](https://github.com/openedx/frontend-app-admin-console/issues/149).
Turning the MFE into a library that runs inside the
frontend-base shell. Follows the official ["Migrating an MFE to frontend-base"](https://github.com/openedx/frontend-base/blob/main/docs/how_tos/migrate-frontend-app.md) guide.

### How I have tested this change?

- Ran unit tests
- Built the project
- Ran the project locally and tested manually the different views

### Screenshots

Screenshot 2026-08-19 at 11 31 06 a m

Screenshot 2026-08-19 at 11 31 28 a m

Screenshot 2026-08-19 at 11 31 39 a m

Closes #211


openedx-webhooks:
Thanks for the pull request, @jesusbalderramawgu!


</details>

- Dependencies
- Blockers
- Timeline information
- Partner information
- Supporting documentation
</details>





</details><details>


---

</details>

- The need for product review

</details>

src/authz-module/components/TableControlBar/types.ts
```diff
@@ -1,19 +1,19 @@
...
-  disabled?: boolean;
+  filterButtonText: string,
+  filterChoices: FilterChoice[],
+  filterValue: string[] | undefined,
+  setFilter: (value: string[], newItem: FilterChoice) => void,
+  isGrouped?: boolean,
+  isSearchable?: boolean,
+  onSearchChange?: (value: string) => void,
+  iconSrc?: React.ComponentType | undefined,
+  disabled?: boolean,
```
bradenmacdonald: What changed these `;` to `,` ? [Semicolons are the standard for TypeScript interfaces](https://www.typescriptlang.org/docs/handbook/2/objects.html) and most code formatters and linters enforce that. Restoring these to `;` is not only more correct but will make this diff smaller.

src/authz-module/components/TableCells.test.tsx
```diff
@@ -59,6 +63,10 @@ describe('TableCells Components', () => {
           userId: 1,
           username: 'testuser',
           email: 'testuser@example.com',
+          name: 'Test User',
+          administrator: false,
+          roles: [],
+          avatar: '',
```
bradenmacdonald: I think this boilerplate should be factored out into a common `testUtils` file. It's very verbose to have it repeated so many times in each test suite. [The way we do it via `testUtils` in the authoring MFE](https://github.com/openedx/frontend-app-authoring/blob/de0e1c43f975d2a643f611783155d99c608c09d6/src/testUtils.tsx#L158-L178) has been nice.

```tsx
    beforeEach(() => {
      initializeMockApp({
        authenticatedUser: {
          userId: 1,
          username: 'testuser',
          email: 'testuser@example.com',
          name: 'Test User',
          administrator: false,
          roles: [],
          avatar: '',
        },
      });
```

src/authz-module/components/TableControlBar/types.ts
jesusbalderramawgu: I changed this because the linter was complaining about the ; and I checked the instructor dashboard and it has commas as well.
This is an example of the instructor dashboard.
Screenshot 2026-08-19 at 11 53 48 a m

src/authz-module/components/TableCells.test.tsx
jesusbalderramawgu: agree! I will refactor this, thanks

src/authz-module/components/TableControlBar/types.ts
bradenmacdonald: Hmm, @arbrandes do you if it was a deliberate decision [to change the eslint config in frontend-base](https://github.com/openedx/frontend-base/commit/8708c076938ee5c4879d9cef93ee83f41a458c3a) so that [`member-delimiter-style`](https://eslint.style/rules/member-delimiter-style) requires commas instead of the default semicolons?

It's not a big deal, but it seems like it's making all our "convert to frontend-base" diffs noisy, and goes against the usual standard in the TypeScript world.

arbrandes: Yeah, I've run into this in the past, and found it annoying.  I don't remember ever discussing this with David, though, so don't actually know why.

But at the same time it feels weird to go back to all the frontend-base repos to change it back.  Ug.

bradenmacdonald: IMHO that's why we should actually use the "warnings vs. errors" feature that linters (eslint & oxlint) have. We could mark the comma as a warning, the semicolon as correct, and then never worry about it again. This diff would be smaller, and the existing frontend-base codebases would be fine, and could be slowly brought into alignment.

But we have a configuration that basically treats warnings as errors, so we don't really take advantage of that.

arbrandes: I'm not a fan of warnings because they just pile up into a mountain of useless noise.  We should just pick one and deal with it.  A one-time set of commits changing one to the other is not the end of the world and pretty easy for an LLM.

@brian-smith-tcril, @diana-villalvazo-wgu, as co-maintainers of frontend-base, thoughts?

arbrandes: Or, better yet, let's discuss this in the [issue](https://github.com/openedx/frontend-base/issues/296) (or in the [PR](https://github.com/openedx/frontend-base/pull/297)).

arbrandes: [Merged](https://github.com/openedx/frontend-base/pull/297) and available as of frontend-base 2.0.0-alpha.2.

jesusbalderramawgu: Thank you, I will update this PR ASAP

jesusbalderramawgu:
I have updated the PR with the latest changes from frontend base.

I rebased this branch and noticed that codecov is complaining, I will check to cover the pending test coverage.
anything please let me know. Thank you!

src/routes.tsx
```diff
@@ -0,0 +1,16 @@
...
+
+const routes = [
+  {
+    id: 'org.openedx.frontend.route.adminConsole.main',
+    path: '/authz/*',
+    Component: Main,
+    handle: {
+      role: 'org.openedx.frontend.role.adminConsole',
+    },
+  },
```
arbrandes: `handle.role` should be `handle.roles`.

src/routes.tsx
```diff
@@ -0,0 +1,16 @@
+import { lazy } from 'react';
+
+const Main = lazy(() => import('./Main'));
+
+const routes = [
+  {
+    id: 'org.openedx.frontend.route.adminConsole.main',
+    path: '/authz/*',
+    Component: Main,
```
arbrandes, line 11: The app no longer requires an authenticated user. Please add `loader: authenticatedLoader` to the route.

src/app.ts
```diff
@@ -0,0 +1,15 @@
...
+import routes from './routes';
+import providers from './providers';
+
+const app: App = {
+  appId,
+  routes,
+  providers,
+  config: {
+    COURSE_AUTHORING_MICROFRONTEND_URL: 'http://apps.local.openedx.io:2001/authoring/home',
+  },
```
arbrandes: Two problems on these three lines.

Let's not ship a local dev URL is as default.

But more importantly, inter-app URLs should become `externalRoutes`. `site.config.dev.tsx` already declares external routes for profile and account, so authoring would sit naturally alongside them.

package.json
```diff
@@ -35,12 +56,10 @@
...
+  "peerDependencies": {
     "@edx/brand": "npm:@openedx/brand-openedx@^1.2.3",
-    "@edx/frontend-component-header": "^8.0.0",
-    "@edx/frontend-platform": "^8.7.0",
     "@edx/openedx-atlas": "^0.7.0",
-    "@openedx/frontend-plugin-framework": "^1.7.0",
+    "@openedx/frontend-base": "^2.0.0-alpha.2 || 0.0.0-dev",
     "@openedx/paragon": "^23.15.1",
     "@tanstack/react-query": "5.96.2",
     "lodash.debounce": "^4.0.8",
```
arbrandes: This renames the entire old `dependencies` block to `peerDependencies`. The howto moves only the deps *shared with the shell*, which pulls four things along that don't belong here.

The ranges are also tighter than peer ranges should be: `@openedx/paragon` `^23.15.1` → `^23`, `react-router-dom` `^6.30.3` → `^6`.

In particular, don't pin `@tanstack/react-query` to an exact version.

src/data/utils.ts
```diff
@@ -1,6 +1,7 @@
-import { getConfig } from '@edx/frontend-platform';
+import { getSiteConfig } from '@openedx/frontend-base';
 
-export const getApiUrl = (path: string) => `${getConfig().LMS_BASE_URL}${path || ''}`;
+export const getApiUrl = (path: string) => `${getSiteConfig().lmsBaseUrl}${path || ''}`;
+export const getStudioApiUrl = (path: string) => `${getSiteConfig().cmsBaseUrl}${path || ''}`;
```
arbrandes: `getStudioApiUrl` is never called anywhere in the repo. `cmsBaseUrl` was added to `site.config.test.tsx:10` only to feed it, and isn't in the dev or CI configs. Worth dropping both from a migration PR.

src/authz-module/data/api.ts
```diff
@@ -32,13 +40,18 @@ export interface DeleteRevokeUserRolesResponse {
...
     error: string;
-  }[],
+  }[];
 }
 
+export interface PermissionsByRole {
+  role: string;
+  permissions: string[];
+  userCount: number;
+};
```
arbrandes: `PermissionsByRole` is unreferenced, as is `GetTeamMembersResponse` added at line 20. Both are new in this PR - worth dropping since they're unrelated to the migration.

Also note the stray `;` after the closing brace on line 54 (see the `src/types.ts` comment for the full list).

tsconfig.build.json
```diff
@@ -0,0 +1,19 @@
...
+  ],
+  "exclude": [
+    "src/**/*.test.ts",
+    "src/**/*.test.tsx",
+    "src/**/*.spec.ts",
+    "src/**/*.spec.tsx",
+    "src/__mocks__/**/*",
+    "src/setupTest.tsx"
+  ]
+}
```
arbrandes, line 20: Please add `src/testUtils.tsx` to `exclude`. The list covers `*.test.*`, `__mocks__` and `setupTest.tsx` but not `testUtils.tsx`, so `npm run build` emits `dist/testUtils.js` (verified locally), which imports `@testing-library/react` - a devDependency - into the shipped tarball.

```suggestion
  "exclude": [
    "src/**/*.test.ts",
    "src/**/*.test.tsx",
    "src/**/*.spec.ts",
    "src/**/*.spec.tsx",
    "src/__mocks__/**/*",
    "src/setupTest.tsx",
    "src/testUtils.tsx"
  ]
}
```

src/testUtils.tsx
```diff
@@ -0,0 +1,106 @@
...
+export function initializeMocks({ user = authenticatedUser } = {}) {
+  initializeMockApp({ authenticatedUser: user });
+}
+
+/** Value provided to `SiteContext` in tests. */
+export const mockAppContext = {
+  authenticatedUser,
+  siteConfig: {
+    ...process.env,
+  } as unknown as SiteConfig,
```
arbrandes: `{ ...process.env }` as the test `siteConfig` is a frontend-platform-era leftover; the howto's "Stop using `process.env`" step covers this. Use `getSiteConfig()` instead - `setupTest.tsx` already calls `mergeSiteConfig(siteConfig)`, so it will return the real test config.

It's inert today because the only `SiteContext` consumers in the app read `authenticatedUser`, but it will silently mislead the first test that reads config through the context.

src/authz-module/components/TableCells.tsx
```diff
@@ -1,5 +1,5 @@
-import { useIntl } from '@edx/frontend-platform/i18n';
-import { AppContext } from '@edx/frontend-platform/react';
+import { useIntl } from '@openedx/frontend-base';
+import { SiteContext } from '@openedx/frontend-base';
```
arbrandes: These two can be one import, and the same pattern shows up in eight files as a byproduct of swapping sub-paths for the root package:

```suggestion
import { SiteContext, useIntl } from '@openedx/frontend-base';
```

The others: `audit-user/index.tsx:5-6`, `components/ToastManager/ToastManagerContext.tsx:4-5`, `components/ErrorPage/index.tsx:3-4`, `authz-module/data/api.ts:1,6`, `data/api.ts:1,3`, `role-assignation-wizard/components/SelectUsersAndRoleStep.tsx:2,7`, `hooks/useCourseAuthoringFlag.test.tsx`.

src/types.ts
```diff
@@ -40,13 +60,13 @@ export type PermissionItem = ResourceMetadata & {
   perms: PermissionMetadata[];
 };
 
-export type Org = {
+export interface Org {
   id: string;
   name: string;
   shortName: string;
 };
```
arbrandes: Stray `;` after an interface body - a leftover from the `type X = {...};` → `interface X {...}` conversion. Lint doesn't flag these, so they need a sweep; there are 17 of them:

- `src/types.ts`: 49, 57, 67, 74, 105
- `src/authz-module/data/api.ts`: 37, 54, 88, 98
- `src/data/types.ts`: 7, 39
- `src/authz-module/components/TableCells.tsx`: 24, 47
- `src/authz-module/components/TableControlBar/types.ts`: 7
- `src/authz-module/components/PermissionTable.tsx`: 16
- `src/authz-module/roles-permissions/utils.ts`: 47
- `src/components/ToastManager/ToastManagerContext.tsx`: 40

src/routes.test.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+    it('has the expected id and path', () => {
+      expect(routes[0].id).toBe('org.openedx.frontend.route.adminConsole.main');
+      expect(routes[0].path).toBe('/authz/*');
+    });
+
+    it('is guarded by the admin console role', () => {
+      expect(routes[0].handle).toEqual({ role: 'org.openedx.frontend.role.adminConsole' });
+    });
+
```
arbrandes: This assertion is what let the singular `role` form survive, so it needs updating alongside the route itself.

The description is also slightly off: route roles don't guard anything, they label a route's purpose. The test that would genuinely be worth having here is one asserting `loader` is set, covering the lost authentication requirement.

README.rst
```diff
@@ -36,9 +36,9 @@ This repository works with `openedx-authz <https://github.com/openedx/openedx-au
 
 Plugins
 =======
-This MFE can be customized using `Frontend Plugin Framework <https://github.com/openedx/frontend-plugin-framework>`_.
+This frontend app can be customized using `Frontend Plugin Framework <https://github.com/openedx/frontend-plugin-framework>`_.
 
-The parts of this MFE that can be customized in that manner are documented `here </src/plugin-slots>`_.
+The parts of this frontend app that can be customized in that manner are documented `here </src/plugin-slots>`_.
```
arbrandes: Two stale references here:

- Line 41 links to `/src/plugin-slots`, which this PR renamed to `src/slots`. Dead link.
- Line 39 still points customization at Frontend Plugin Framework, which this PR uninstalls. Slots are now frontend-base's `<Slot />`, per the [slot naming ADR](https://github.com/openedx/frontend-base/blob/main/docs/decisions/0009-slot-naming-and-lifecycle.rst).

While in the area: `src/slots/README.md` is still titled "Plugin Slots" and has no body, and `docs/how_tos/i18n.rst` forwards to frontend-platform's i18n howto.

Makefile
```diff
@@ -1,44 +1,74 @@
...
-transifex_utils = ./node_modules/.bin/transifex-utils.js
-i18n = ./src/i18n
-transifex_input = $(i18n)/transifex_input.json
+npm-install-%: ## install specified % npm package
+	npm install $* --save-dev
+	git add package.json
 
-# This directory must match .babelrc .
-transifex_temp = ./temp/babel-plugin-formatjs
+TURBO = TURBO_TELEMETRY_DISABLED=1 turbo --dangerously-disable-package-manager-check
```
arbrandes: `--dangerously-disable-package-manager-check` isn't necessary. Add `"packageManager": "npm@11.12.1"` to `package.json` and drop the flag:

```suggestion
TURBO = TURBO_TELEMETRY_DISABLED=1 turbo
```

Makefile
```diff
@@ -1,44 +1,74 @@
...
+	$(MAKE) bin-link
 
-i18n.concat:
-	# Gathering JSON messages into one file...
-	$(transifex_utils) $(transifex_temp) $(transifex_input)
+clean-packages: turbo.json
+	$(TURBO) run clean; rm -f turbo.json
 
-extract_translations: | requirements i18n.extract i18n.concat
+dev-packages: turbo.json
```
arbrandes: Missing the `build-packages` prerequisite, so a fresh `npm run dev:packages` starts the watch without having bin-linked frontend-base first:

```suggestion
dev-packages: build-packages turbo.json
```

`dev-site` does call `bin-link`, so this may work in practice, but the howto and `frontend-app-catalog/Makefile:29` both have the prerequisite.

Makefile
```diff
@@ -1,44 +1,74 @@
...
-	git diff --exit-code $(i18n)
+dev-site: bin-link
+	npm run dev
+
+clean:
+	rm -rf dist
+
+build:
+	tsc --project tsconfig.build.json
+	find src -type f \( -name '*.scss' -o \( \( -name '*.png' -o -name '*.svg' \) -path '*/assets/*' \) \) -exec sh -c '\
```
arbrandes: This enumerates `*.png`/`*.svg` under `assets/`, where the howto copies everything under `assets/` regardless of type. Any other asset type - a font, a `.webp`, a `.json` fixture - will silently not ship in `dist/`. The howto's form is:

```
	find src -type f \( -name '*.scss' -o -path '*/assets/*' \) -exec sh -c '\
```

site.config.test.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+const siteConfig: SiteConfig = {
+  siteId: 'admin-console-test-site',
+  siteName: 'Admin Console Test Site',
+  baseUrl: 'http://localhost:2025',
+  lmsBaseUrl: 'http://localhost:8000',
+  cmsBaseUrl: 'http://studio.local.openedx.io:8001',
+  loginUrl: 'http://localhost:8000/login',
+  logoutUrl: 'http://localhost:8000/logout',
+  environment: EnvironmentTypes?.TEST ?? 'test',
```
arbrandes: Prefer the documented form here:

```suggestion
  // Use 'test' instead of EnvironmentTypes.TEST to break a circular dependency
  // when mocking `@openedx/frontend-base` itself.
  environment: 'test' as SiteConfig['environment'],
```

The optional chain is guarding against the same circular-mock problem, but non-obviously - a reader can't tell why an imported enum member might be undefined.

Note this leaves `EnvironmentTypes` unused on line 1, so that import needs trimming to `import type { SiteConfig } from '@openedx/frontend-base';` or lint will fail.

.npmignore
```diff
@@ -1,12 +1,6 @@
...
-npm-debug.log
-
-coverage
+__mocks__
 node_modules
-public
+*.test.js
+*.test.jsx
+*.test.ts
+*.test.tsx
```
arbrandes: Since `files: ["/dist"]` in `package.json` already whitelists what ships, the howto trims this to just:

```suggestion
node_modules
```

.gitignore
```diff
@@ -1,19 +1,21 @@
-.DS_Store
-.eslintcache
-.idea
 node_modules
 npm-debug.log
 coverage
 module.config.js
```
arbrandes: `module.config.js` is a frontend-build concept superseded by npm workspaces, so this line and the `module.config.js.example` file at the repo root can both go.

arbrandes (changes requested):

package.json
```diff
@@ -1,32 +1,53 @@
 {
   "name": "@openedx/frontend-app-admin-console",
-  "version": "0.1.0",
+  "version": "1.0.0",
```
arbrandes:
```suggestion
  "version": "0.0.0-dev",
```

arbrandes:
Also, please rebase on latest master so we don't lose any new features/fixes.  I see a couple landed, recently.

jesusbalderramawgu:
Thank you @arbrandes for your great feedback!
I have addressed your comments, anything please let me know!

src/authz-module/components/ErrorPage/index.tsx
```diff
@@ -70,23 +70,23 @@ const ErrorPage = ({ error, resetErrorBoundary }: FallbackProps) => {
...
-          as={Hyperlink}
-          destination={`${getConfig().COURSE_AUTHORING_MICROFRONTEND_URL}`}
-          className="m-2"
-          variant={showReloadButton ? 'outline-primary' : 'primary'}
-        >
-          {intl.formatMessage(messages['error.page.action.back'])}
-        </Button>
+          <Button
+            as={Hyperlink}
+            destination={`${getAppConfig(appId).COURSE_AUTHORING_MICROFRONTEND_URL}`}
```
arbrandes: Use `destination={getUrlByRouteRole('org.openedx.frontend.role.courseAuthoring')}`, drop the `getAppConfig`/`appId` imports, and remove the now-dead `COURSE_AUTHORING_MICROFRONTEND_URL` entry in `site.config.test.tsx`.

tsconfig.build.json
```diff
@@ -0,0 +1,19 @@
...
+    "src/**/*"
+  ],
+  "exclude": [
+    "src/**/*.test.ts",
+    "src/**/*.test.tsx",
+    "src/**/*.spec.ts",
+    "src/**/*.spec.tsx",
+    "src/__mocks__/**/*",
+    "src/setupTest.tsx"
+  ]
```
arbrandes, line 19: Add `"src/testUtils.tsx"` here, and delete the stray file named `tsconfig.build.json ` (with a trailing space) that the exclusion landed in instead.

src/authz-module/audit-user/index.test.tsx
```diff
@@ -1,31 +1,23 @@
...
   render, screen, waitFor, act,
 } from '@testing-library/react';
-import { AppContext } from '@edx/frontend-platform/react';
+import { SiteContext } from '@openedx/frontend-base';
 import userEvent from '@testing-library/user-event';
 import { MemoryRouter, Route, Routes } from 'react-router-dom';
-import { mockHttpClient, mockAppContext } from '@src/setupTest';
-import { IntlProvider } from '@edx/frontend-platform/i18n';
+import { mockHttpClient, mockAppContext } from '@src/testUtils';
+import { IntlProvider } from '@openedx/frontend-base';
```
arbrandes: Merge these two `@openedx/frontend-base` imports. They're non-adjacent, so the earlier sweep missed them; this is the last file with two.

src/slots/README.md
```diff
@@ -0,0 +1,5 @@
+# `frontend-app-admin-console` Slots
+
+This directory contains the slots that this frontend app _offers_ for host sites to customize. Slots are implemented with the `<Slot />` component from [`@openedx/frontend-base`](https://github.com/openedx/frontend-base); see the [slot naming and lifecycle ADR](https://github.com/openedx/frontend-base/blob/main/docs/decisions/0009-slot-naming-and-lifecycle.rst) for the API and conventions.
+
+Slots that this app _uses_ from other packages are declared in `src/slots.tsx`.
```
arbrandes: Drop this line - `src/slots.tsx` doesn't exist, so it points readers at a dead path.

package.json
```diff
@@ -35,30 +57,37 @@
...
+    "nodemon": "^3.1.14",
+    "ts-jest": "^29.4.5",
+    "tsc-alias": "^1.9.1",
+    "turbo": "^2.10.11"
+  },
+  "overrides": {
+    "react-intl": "6.8.9"
+  },
+  "dependencies": {
+    "@edx/brand": "npm:@openedx/brand-openedx@^1.2.3",
```
arbrandes: Remove `@edx/brand` - nothing in the repo imports it (the dev harness gets branding via `@openedx/frontend-base/shell/style`), so as a hard dependency it just makes every consumer install a brand alias they don't use. `lodash.debounce` and `react-error-boundary` moving here was right.

package.json
```diff
@@ -35,30 +57,37 @@
...
     "@testing-library/jest-dom": "^6.6.4",
     "@testing-library/react": "^16.3.0",
     "@testing-library/user-event": "^14.6.1",
+    "@types/jest": "^29.5.14",
     "@types/react": "^18",
     "@types/react-dom": "^18",
-    "ts-jest": "^29.4.5"
+    "@types/react-helmet": "^6.1.11",
+    "nodemon": "^3.1.14",
+    "ts-jest": "^29.4.5",
```
arbrandes: `ts-jest` isn't referenced by `jest.config.js` or `babel.config.js`, and neither `frontend-app-catalog` nor `frontend-app-instructor-dashboard` carries it.

package.json
```diff
@@ -35,30 +57,37 @@
...
+    "tsc-alias": "^1.9.1",
+    "turbo": "^2.10.11"
+  },
+  "overrides": {
+    "react-intl": "6.8.9"
+  },
```
arbrandes: This should go away, and the repo fixed up for the version frontend-base provides.

arbrandes (changes requested):
Thanks!  Next round!

jesusbalderramawgu:
> Thanks! Next round!

comments addressed! thank you!

src/authz-module/role-assignation-wizard/components/SelectUsersAndRoleStep.test.tsx
```diff
@@ -1,9 +1,10 @@
 import { screen } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
-import { renderWrapper } from '@src/setupTest';
+import { renderWrapper } from '@src/testUtils';
 import SelectUsersAndRoleStep from './SelectUsersAndRoleStep';
 
-jest.mock('@edx/frontend-platform', () => ({
+jest.mock('@openedx/frontend-base', () => ({
+  ...jest.requireActual('@openedx/frontend-base'),
```
dcoa: `getConfig` is dead code here 👇

site.config.dev.tsx
```diff
@@ -0,0 +1,47 @@
...
+      role: 'org.openedx.frontend.role.courseAuthoring',
+      url: 'http://apps.local.openedx.io:2001/authoring/home'
+    },
+    {
+      role: 'org.openedx.frontend.role.logout',
+      url: 'http://local.openedx.io:8000/logout'
+    },
+  ],
+
+  accessTokenCookieName: 'edx-jwt-cookie-header-payload', // check this
```
dcoa: it is the comment needed?

src/style.scss
dcoa, line 1: Do we need this file?

**Note:** I just noticed some toast styles that might be better here

src/providers.ts
```diff
@@ -0,0 +1,5 @@
+import { AppProvider } from '@openedx/frontend-base';
+
+const providers: AppProvider[] = [];
```
dcoa: I think a good improvement here is adding the `ToastProvider` here, it is currently under authz module but it is an App level component. It will be useful for other modules too.


```suggestion
const providers: AppProvider[] = [
  ToastManagerProvider,
];

dcoa (commented):
I just find a couple of nits

jesusbalderramawgu:
@dcoa  Thank you for your feedback, comments have been addressed

dcoa:
Thank you, @jesusbalderramawgu, for addressing my comments. I’ve left the rest of the migration review to @arbrandes, as he is more familiar with it.

arbrandes (approved):
Looks great, thanks!

I'm going to rebase/merge this myself as per the procedure outlined in https://github.com/openedx/frontend-base/issues/243, which means I'm approving the PR (congrats!) but also closing it.

arbrandes:
The conversion has landed on `master` via #227: the reviewed work here was squashed into a single commit preserving @jesusbalderramawgu's authorship, rebased onto `master`'s tip, and fast-forwarded per [OEP-65 ADR 0004](https://docs.openedx.org/projects/openedx-proposals/en/latest/architectural-decisions/oep-0065/decisions/0004-merging-frontend-base.html). Closing as superseded by #227. Thank you for the migration!

